PR #18, opened 2022-12-21
chore: get subsystem with nqn to list all subsystems for volume

abhilashshetty04:
This PR adds method to get all subsystems registered with the supplied nqn.

Signed-off-by: Abhilash Shetty <abhilash.shetty@datacore.com>

### DIFF
--- a/nvmeadm/src/error.rs
+++ b/nvmeadm/src/error.rs
@@ -19,8 +19,8 @@ pub enum NvmeError {
         filename: String,
         source: std::io::Error,
     },
-    #[snafu(display("nqn: {} not found", text))]
-    NqnNotFound { text: String },
+    #[snafu(display("No Nvmf subsystem found for the nqn: {}", nqn))]
+    NqnNotFound { nqn: String },
     #[snafu(display("No nvmf subsystems found"))]
     NoSubsystems,
     #[snafu(display(
@@ -34,8 +34,6 @@ pub enum NvmeError {
         host: String,
         port: u16,
     },
-    #[snafu(display("No Nvmf Subsystem found for nqn :{}", nqn))]
-    NoSubsytemFound { nqn: String },
     #[snafu(display("Connect in progress"))]
     ConnectInProgress,
     #[snafu(display("NVMe connect failed: {}, {}", filename, source))]
--- a/nvmeadm/src/error.rs
+++ b/nvmeadm/src/error.rs
@@ -34,6 +34,8 @@ pub enum NvmeError {
         host: String,
         port: u16,
     },
+    #[snafu(display("No Nvmf Subsystem found for nqn :{}", nqn))]
+    NoSubsytemFound { nqn: String },
     #[snafu(display("Connect in progress"))]
     ConnectInProgress,
     #[snafu(display("NVMe connect failed: {}, {}", filename, source))]
--- a/nvmeadm/src/nvmf_subsystem.rs
+++ b/nvmeadm/src/nvmf_subsystem.rs
@@ -172,6 +172,24 @@ impl Subsystem {
             Some(subsys) => Ok(subsys),
         }
     }
+
+    /// Gets all Nvme subsystems registered for a given volume.
+    pub fn get_from_nqn(nqn: &str) -> Result<Vec<Subsystem>, NvmeError> {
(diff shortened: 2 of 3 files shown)


nvmeadm/src/nvmf_subsystem.rs
```diff
@@ -172,6 +172,24 @@ impl Subsystem {
             Some(subsys) => Ok(subsys),
         }
     }
+
+    /// Gets all Nvme subsystems registered for a given volume.
+    pub fn get_from_nqn(nqn: &str) -> Result<Vec<Subsystem>, NvmeError> {
```
tiagolobocastro:
```suggestion
    pub fn try_from_nqn(nqn: &str) -> Result<Vec<Subsystem>, NvmeError> {
```

nvmeadm/src/nvmf_subsystem.rs
```diff
@@ -172,6 +172,24 @@ impl Subsystem {
             Some(subsys) => Ok(subsys),
         }
     }
+
+    /// Gets all Nvme subsystems registered for a given volume.
```
tiagolobocastro:
```suggestion
    /// Gets all Nvme subsystems registered for a given nqn.
```

nvmeadm/src/nvmf_subsystem.rs
```diff
@@ -172,6 +172,24 @@ impl Subsystem {
...
+    pub fn get_from_nqn(nqn: &str) -> Result<Vec<Subsystem>, NvmeError> {
+        let nvme_subsystems = NvmeSubsystems::new()?;
+        let mut nvme_paths = vec![];
+        for path in nvme_subsystems.flatten() {
+            if path.nqn == nqn && (path.state == "live" || path.state == "connecting") {
```
tiagolobocastro: Should we split this into 2 fns, one that gets matching non subsystems, and another than filters it down to only live or connecting paths?

nvmeadm/src/nvmf_subsystem.rs
```diff
@@ -172,6 +172,24 @@ impl Subsystem {
...
+                nvme_paths.push(path);
+            }
+        }
+        if !nvme_paths.is_empty() {
```
tiagolobocastro: isn't this logic reversed?

nvmeadm/src/error.rs
```diff
@@ -34,6 +34,8 @@ pub enum NvmeError {
         host: String,
         port: u16,
     },
+    #[snafu(display("No Nvmf Subsystem found for nqn :{}", nqn))]
+    NoSubsytemFound { nqn: String },
```
tiagolobocastro: We already have an `NqnNotFound` error, could we reuse it?

nvmeadm/src/nvmf_subsystem.rs
tiagolobocastro: Maybe the filtering for live or connecting can be done on the control-plane repo

nvmeadm/src/error.rs
abhilashshetty04: I thought that error is not descriptive enough. Made minor change in formating and using it.

nvmeadm/src/nvmf_subsystem.rs
abhilashshetty04: Incorporated

nvmeadm/src/nvmf_subsystem.rs
abhilashshetty04: Yes it was wrong. Fixed it

nvmeadm/src/error.rs
abhilashshetty04: Incorporated

tiagolobocastro: Then you can fix that error and make it more descriptive, win win :)

tiagolobocastro: Oh I see that's what you did already, great! 👍

nvmeadm/src/nvmf_subsystem.rs
abhilashshetty04: Incorporated

tiagolobocastro (approved):
LGTM

PR #25, opened 2023-04-25
feat(composer): adding method to print logs for the given container

dsavitskiy:


composer/src/composer.rs
```diff
@@ -1695,4 +1690,21 @@ impl ComposeTest {
         let exit_code = self.docker.inspect_exec(&exec.id).await?.exit_code;
         Ok((exit_code, response))
     }
+
+    /// Print logs for all containers.
+    pub fn print_all_logs(&self) {
+        self.containers.keys().for_each(|name| {
+            self.print_log(name);
+            tracing::error!("");
```
tiagolobocastro: Do we need this extra log?

tiagolobocastro: If it's for padding, then maybe would be better a println! ?

dsavitskiy: yes indeed

tiagolobocastro (approved):

PR #29, opened 2023-05-24
feat(eventing): add Nats mbus api

datacore-vvarakantham:
This PR contains Nats Message bus module. Consists of connection, publish, subscribe functions for the NATS

### DIFF
--- a/mbus-api/Cargo.toml
+++ b/mbus-api/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "mbus-api"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+async-nats = "0.29.0"
+bytes = "1.4.0"
+serde = "1.0.152"
+serde_json = "1.0.93"
+tokio = "1.25.0"
+tracing = "0.1.37"
+uuid = { version = "1.3.0", features = ["v4"] }
+async-trait = "0.1.64"
+futures = "0.3.26"
+snafu = "0.7.4"
--- a/mbus-api/src/common/constants.rs
+++ b/mbus-api/src/common/constants.rs
@@ -1,0 +1,23 @@
+use std::time::Duration;
+
+/// Each event message is nearly of size 0.3 KB. So the stream of this size(3 MB) can hold nearly
+/// 10K messages.
+pub const STREAM_SIZE: i64 = 3 * 1024 * 1024;
+
+/// Stream name for the events.
+pub const STREAM_NAME: &str = "events-stream";
+
+/// Stats consumer name for message bus.
+pub const CONSUMER_NAME: &str = "stats-events-consumer";
+
+/// Subjects for events stream.
+pub const SUBJECTS: &str = "events.>";
+
+/// Timeout for jetstream publish.
+pub const PUBLISH_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// Replica count for messages. Maximum 5.
+pub const NUM_STREAM_REPLICAS: usize = 3;
+
+/// Max msgs per subject.
+pub const MAX_MSGS_PER_SUBJECT: i64 = 1;
--- a/mbus-api/src/common/retry.rs
+++ b/mbus-api/src/common/retry.rs
@@ -1,0 +1,91 @@
+use std::{cmp::min, time::Duration};
+
+/// Backoff delay options for the retries to the NATS jetstream message bus.
+/// Max number of retries until it gives up.
+#[derive(Clone, Debug)]
+pub struct BackoffOptions {
+    /// Initial delay.
+    pub init_delay: Duration,
+    /// The number of attempts with initial delay.
+    pub cutoff: u32,
+    /// Increase in delay with each retry after cutoff is reached.
+    pub step: Duration,
+    /// Maximum delay.
+    pub max_delay: Duration,
+    /// Maximum retries.
+    pub max_retries: u32,
+}
+
+impl Default for BackoffOptions {
+    fn default() -> Self {
+        Self {
+            init_delay: Duration::from_secs(5),
+            cutoff: 4,
+            step: Duration::from_secs(2),
+            max_delay: Duration::from_secs(10),
+            max_retries: 10,
+        }
+    }
+}
+
+impl BackoffOptions {
+    /// New options with default values.
+    #[must_use]
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Initial delay before the first retry.
+    pub fn with_init_delay(mut self, init_delay: Duration) -> Self {
+        self.init_delay = init_delay;
+        self
+    }
+
+    /// Delay multiplied at each iteration.
+    pub fn with_delay_step(mut self, step: Duration) -> Self {
+        self.step = step;
+        self
+    }
+
+    /// Number of tries with the initial delay.
+    pub fn with_cutoff(mut self, cutoff: u32) -> Self {
+        self.cutoff = cutoff;
+        self
+    }
+
+    /// Maximum delay.
+    pub fn with_max_delay(mut self, max_delay: Duration) -> Self {
+        self.max_delay = max_delay;
+        self
+    }
+
+    /// Specify a max number of retries before giving up.
+    pub fn with_max_retries(mut self, max_retries: u32) -> Self {
+        self.max_retries = max_retries;
+        self
+    }
+
+    pub fn publish_backoff_options() -> Self {
+        Self {
+            init_delay: Duration::from_secs(5),
+            cutoff: 4,
+            step: Duration::from_secs(2),
+            max_delay: Duration::from_secs(10),
+            max_retries: 10,
+        }
+    }
+}
+
+/// Simple backoff delay which get gradually larger up to a 'max' duration.
+pub async fn backoff_with_options(tries: &mut u32, options: &BackoffOptions) {
+    *tries += 1;
+    let backoff = if *tries <= options.cutoff {
+        options.init_delay
+    } else {
+        min(
+            options.init_delay + (*tries - options.cutoff - 1) * options.step,
+            options.max_delay,
+        )
+    };
+    tokio::time::sleep(backoff).await;
+}
--- a/mbus-api/src/lib.rs
+++ b/mbus-api/src/lib.rs
@@ -1,0 +1,18 @@
+use async_trait::async_trait;
+use mbus_nats::{BusResult, BusSubscription};
+use message::EventMessage;
+use serde::{de::DeserializeOwned, Serialize};
+
+mod common;
+pub mod mbus_nats;
+pub mod message;
+
+#[async_trait]
+pub trait Bus {
+    /// Publish a message to message bus.
+    async fn publish(&mut self, message: &EventMessage) -> BusResult<u64>;
+    /// Create a subscription which can be
+    /// polled for messages until the bus is closed.
+    async fn subscribe<T: Serialize + DeserializeOwned>(&mut self)
+        -> BusResult<BusSubscription<T>>;
+}
--- a/mbus-api/src/mbus_nats.rs
+++ b/mbus-api/src/mbus_nats.rs
@@ -1,0 +1,337 @@
+use crate::{
+    common::{
+        constants::*,
+        errors::Error,
+        retry::{backoff_with_options, BackoffOptions},
+    },
+    message::EventMessage,
+    Bus,
+};
+use async_nats::{
+    jetstream::{
+        self,
+        consumer::{
+            push::{Config, Messages},
+            DeliverPolicy,
+        },
+        stream::Stream,
+        Context,
+    },
+    Client,
+};
+use async_trait::async_trait;
+use bytes::Bytes;
+use futures::StreamExt;
+use serde::{de::DeserializeOwned, Serialize};
+use std::{io::ErrorKind, marker::PhantomData, time::Duration};
+
+/// Result wrapper for Jetstream requests.
+pub type BusResult<T> = Result<T, Error>;
+
+/// Initialise the Nats Message Bus.
+pub async fn message_bus_init(server: &str) -> NatsMessageBus {
+    NatsMessageBus::new(server).await
+}
+
+/// Nats implementation of the Bus.
+#[derive(Clone)]
+pub struct NatsMessageBus {
+    client: Client,
+    jetstream: Context,
+}
+
+impl NatsMessageBus {
+    /// Connect to nats server.
+    pub async fn connect(server: &str) -> Client {
+        tracing::debug!("Connecting to the nats server {}...", server);
+        // We retry in a loop until successful. Once connected the nats library will handle
+        // reconnections for us.
+        let options = BackoffOptions::new()
+            .with_init_delay(Duration::from_secs(5))
+            .with_cutoff(4)
+            .with_delay_step(Duration::from_secs(2))
+            .with_max_delay(Duration::from_secs(10));
+        let mut tries = 0;
+        let mut log_error = true;
+        loop {
+            match async_nats::ConnectOptions::new()
+                .event_callback(|event| async move {
+                    match event {
+                        async_nats::Event::Disconnected => {
+                            tracing::warn!("NATS connection has been lost")
+                        }
+                        async_nats::Event::Connected => {
+                            tracing::debug!("NATS connection has been reestablished")
+                        }
+                        _ => (),
+                    }
+                })
+                .connect(server)
+                .await
+            {
+                Ok(client) => {
+                    tracing::debug!("Connected to the nats server {}", server);
+                    return client;
+                }
+                Err(error) => {
+                    if log_error {
+                        tracing::warn!(%error, "Nats connection error. Retrying...");
+                        log_error = false;
+                    }
+                    backoff_with_options(&mut tries, &options).await;
+                }
+            }
+        }
+    }
+
+    /// Creates a new nats message bus connection.
+    pub async fn new(server: &str) -> Self {
+        let client = Self::connect(server).await;
+        Self {
+            client: client.clone(),
+            jetstream: {
+                let mut js = jetstream::new(client);
+                js.set_timeout(PUBLISH_TIMEOUT);
+                js
+            },
+        }
+    }
+
+    /// Ensures that the stream is created on jetstream.
+    pub async fn verify_stream_exists(&mut self) -> BusResult<()> {
+        let options = BackoffOptions::new().with_max_retries(0);
+        if let Err(err) = self.get_or_create_stream(Some(options)).await {
+            tracing::warn!(%err,
+                "Error while getting/creating stream '{}'",
+                STREAM_NAME
+            );
+        }
+        Ok(())
+    }
+
+    /// Creates consumer and returns an iterator for the messages on the stream.
+    async fn create_consumer_and_get_messages(&mut self, stream: Stream) -> BusResult<Messages> {
+        tracing::debug!("Getting/creating consumer for stats '{}'", CONSUMER_NAME);
+        let options = BackoffOptions::new();
+        let mut tries = 0;
+        let mut log_error = true;
+
+        let consumer_config = Config {
+            durable_name: Some(CONSUMER_NAME.to_string()),
+            deliver_policy: DeliverPolicy::All,
+            deliver_subject: self.client.new_inbox(),
+            max_ack_pending: 1, /* NOTE: when this value is 1, we receive and ack each message
+                                 * before moving on to the next ordered message */
+            ..Default::default()
+        };
+
+        loop {
+            let err = match stream
+                .get_or_create_consumer(CONSUMER_NAME, consumer_config.clone())
+                .await
+            {
+                Ok(consumer) => match consumer.messages().await {
+                    Ok(messages) => {
+                        tracing::debug!(
+                            "Getting/creating consumer for stats '{}' successful",
+                            CONSUMER_NAME
+                        );
+                        return Ok(messages);
+                    }
+                    Err(error) => error,
+                },
+                Err(error) => error,
+            };
+
+            if tries == options.max_retries {
+                return Err(Error::ConsumerError {
+                    consumer: CONSUMER_NAME.to_string(),
+                    error: err.to_string(),
+                });
+            }
+            if log_error {
+                tracing::warn!(%err,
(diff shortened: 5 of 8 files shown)


mbus-api/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+async-nats = "0.28.0"
+bytes = "1.4.0"
+serde = "1.0.152"
+serde_json = "1.0.93"
+tokio = "1.25.0"
+tracing = "0.1.37"
+uuid = { version = "1.3.0", features = ["v4"] }
+async-trait = "0.1.64"
+futures = "0.3.26"
+snafu = "0.7.4"
```
sinhaashish: add a blank line at the bottom

mbus-api/src/common/constants.rs
```diff
@@ -0,0 +1,23 @@
+use std::time::Duration;
+
+/// Each event message is nearly of size 0.3 KB. So the stream of this size(3 MB) can hold nearly 10
```
tiagolobocastro: doc comments should end with `.`

mbus-api/src/common/constants.rs
```diff
@@ -0,0 +1,23 @@
...
+pub const STREAM_NAME: &str = "events-stream";
+
+/// Stats consumer name for message bus
+pub const CONSUMER_NAME: &str = "stats-events-consumer";
+
+/// Subjects for events stream
+pub const SUBJECTS: &str = "events.*";
+
+/// 3 mins 0 secs - Duplicates are tracked within this window (to publish messages atmost once)
+pub const DUPLICATE_WINDOW: i64 = 180000000000;
```
tiagolobocastro: what is this value in terms of?

mbus-api/src/common/retry.rs
```diff
@@ -0,0 +1,91 @@
+use std::{cmp::min, time::Duration};
+
+/// Backoff delay options for the retries to the NATS jetstream message bus
+/// Max number of retries until it gives up.
+#[derive(Clone, Debug)]
+pub struct BackoffOptions {
+    // initial delay
```
tiagolobocastro: Doc comments should be start with a `C`apital, and end with a `.`

mbus-api/src/common/retry.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl Default for BackoffOptions {
+    fn default() -> Self {
+        Self {
+            init_delay: Duration::from_secs(5),
+            cutoff: 4,
+            step: Duration::from_secs(2),
+            max_delay: Duration::from_secs(10),
+            max_retries: 10,
```
tiagolobocastro, line 26: hmm seems rather low, only 10 retries.
Also, any interaction with the buffer size? Eg: if our buffer is not full, why not keep retrying?

mbus-api/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+pub trait NatsMessage: Send + Sync {
+    /// Returns the subject for the message
+    fn subject(&self) -> String;
+    ///Returns the payload to publish to the message bus
+    fn payload(&self) -> Bytes;
+    ///Returns the headers with the message id to publish to the message bus.
+    /// The message id is required to publish the message atmost once
+    fn headers(&self) -> async_nats::header::HeaderMap;
+    ///Returns the message for logging
+    fn msg(&self) -> String;
```
tiagolobocastro: what is this?

mbus-api/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+use async_trait::async_trait;
+use bytes::Bytes;
+use mbus_nats::{BusResult, BusSubscription};
+use serde::{de::DeserializeOwned, Serialize};
+
+mod common;
+pub mod mbus_nats;
+pub mod message;
+
+pub trait NatsMessage: Send + Sync {
```
tiagolobocastro: Do we really want to tie ourselves to Nats? What about making this generic, in case we need to replace nats with something else?

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,330 @@
...
+                        async_nats::Event::Disconnected => {
+                            tracing::warn!("NATS connection has been lost")
+                        }
+                        async_nats::Event::Connected => {
+                            tracing::info!("NATS connection has been reestablished")
+                        }
+                        _ => (),
+                    }
+                })
+                .connect(server)
```
tiagolobocastro, line 69: is there a lazy connect available so we don't even have to handle initial connection?

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,330 @@
...
+                            CONSUMER_NAME
+                        );
+                        return Ok(messages);
+                    }
+                    Err(error) => error,
+                },
+                Err(error) => error,
+            };
+
+            if tries == options.clone().max_retries {
```
tiagolobocastro: why do we need to clone?

mbus-api/src/common/retry.rs
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+}
+
+/// Simple backoff delay which get gradually larger up to a 'max' duration
+pub async fn backoff_with_options(tries: &mut u32, options: BackoffOptions) {
```
tiagolobocastro: Can we take a reference to the BackoffOptions?

mbus-api/src/message.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub node: String,
+}
+
+impl NatsMessage for EventMessage {
+    fn subject(&self) -> String {
+        format!("events.{}", self.category) // if category is volume, then the subject
+                                            // for the message is 'events.volume'
+    }
+    fn payload(&self) -> bytes::Bytes {
+        Bytes::from(serde_json::to_vec(self).unwrap())
```
tiagolobocastro: Should handle this unwrap by having return type result

mbus-api/src/common/constants.rs
datacore-vvarakantham: Nano seconds. Will add it in the comment.

mbus-api/src/common/constants.rs
```diff
@@ -0,0 +1,23 @@
...
+
+/// Timeout for jetstream publish
+pub const PUBLISH_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// How many replicas to keep for each message in a clustered JetStream, maximum 5
```
sinhaashish: reframe the comment : ` Replica count for messages`.

mbus-api/src/common/retry.rs
```diff
@@ -0,0 +1,91 @@
+use std::{cmp::min, time::Duration};
+
+/// Backoff delay options for the retries to the NATS jetstream message bus
+/// Max number of retries until it gives up.
+#[derive(Clone, Debug)]
+pub struct BackoffOptions {
+    // initial delay
+    pub init_delay: Duration,
```
sinhaashish, line 8: since the variable are public, the comments must have 3 backslashes, so that it converts to doc comments.

mbus-api/src/common/retry.rs
datacore-vvarakantham: Perhaps we can retry the publish from the publisher library when the buffer is <50% filled? We need to increase the duplicate window size too if we increase the number of retries for publish.

mbus-api/src/mbus_nats.rs
datacore-vvarakantham: I didn't find any lazy connect option in the docs. Just the connect returns a client which is a handle to connection.

mbus-api/src/common/retry.rs
tiagolobocastro: Yeah maybe something like that sounds good

mbus-api/src/common/constants.rs
```diff
@@ -0,0 +1,23 @@
+use std::time::Duration;
+
+/// Each event message is nearly of size 0.3 KB. So the stream of this size(3 MB) can hold nearly 10
+/// K messages
```
sinhaashish: 10K must be a single word.

mbus-api/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// Returns the subject for the message
+    fn subject(&self) -> String;
+    ///Returns the payload to publish to the message bus
```
sinhaashish: `///Returns` -> `/// Returns` . Space missing.

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,330 @@
...
+            {
+                Ok(consumer) => match consumer.messages().await {
+                    Ok(messages) => {
+                        tracing::info!(
+                            "Getting/creating consumer for stats '{}' successful",
+                            CONSUMER_NAME
+                        );
+                        return Ok(messages);
+                    }
+                    Err(error) => error,
```
sinhaashish, line 141: Return  specific error using Snafu, this way it will be easy for debugging in case of errors.

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,330 @@
...
+            let err = match self
+                .jetstream
+                .get_or_create_stream(stream_config.clone())
+                .await
+            {
+                Ok(stream) => {
+                    tracing::info!("Getting/creating stream '{}' successful", STREAM_NAME);
+                    return Ok(stream);
+                }
+                Err(error) => error,
```
sinhaashish, line 194: Return  specific error using Snafu, this way it will be easy for debugging in case of errors.

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,330 @@
...
+#[async_trait]
+impl Bus for NatsMessageBus {
+    /// Publish messages to the message bus atmost once. The message is discarded if there is any
+    /// connection error after some retries
+    async fn publish<T: NatsMessage>(&mut self, message: &T) -> BusResult<u64> {
+        let options = BackoffOptions::publish_backoff_options();
+        let mut tries = 0;
+        let mut log_error = true;
+
+        let subject = message.subject().clone();
```
sinhaashish: redundant variable declaration, this can be directly used where required.

mbus-api/src/message.rs
```diff
@@ -0,0 +1,64 @@
+use crate::NatsMessage;
+use bytes::Bytes;
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+// General structure of message bus event
+#[derive(Serialize, Deserialize, Debug)]
+pub struct EventMessage {
+    // Event Category
```
sinhaashish: since it is public use /// for comments.

sinhaashish (commented):
left few more comments

sinhaashish (commented):
left few more comments

sinhaashish (commented):
left few more comments

sinhaashish (commented):
left few more comments

mbus-api/src/mbus_nats.rs
datacore-vvarakantham: The error has the error message from Nats which can be used for debugging. Should I still separate the errors?

mbus-api/src/common/constants.rs
datacore-vvarakantham: Currently I changed the code to remove the duplicate window. For each event msg I am using a unique subject and setting the MAX_MSGS_PER_SUBJECT to 1 in the stream config during stream creation. Which ensures the msgs to be published atmost once with any number of retries. [Infinite msg deduplication](https://nats.io/blog/new-per-subject-discard-policy/) is the link for reference.

mbus-api/src/lib.rs
datacore-vvarakantham: It just returns the msg for logs. But it is not required. I removed it.

sinhaashish (approved):

mbus-api/src/common/constants.rs
```diff
@@ -0,0 +1,23 @@
...
+pub const STREAM_NAME: &str = "events-stream";
+
+/// Stats consumer name for message bus.
+pub const CONSUMER_NAME: &str = "stats-events-consumer";
+
+/// Subjects for events stream.
+pub const SUBJECTS: &str = "events.>";
+
+/// Timeout for jetstream publish.
+pub const PUBLISH_TIMEOUT: Duration = Duration::from_secs(5);
```
tiagolobocastro: Is this perhaps a bit too small?

mbus-api/src/common/constants.rs
```diff
@@ -0,0 +1,23 @@
...
+pub const STREAM_SIZE: i64 = 3 * 1024 * 1024;
+
+/// Stream name for the events.
+pub const STREAM_NAME: &str = "events-stream";
+
+/// Stats consumer name for message bus.
+pub const CONSUMER_NAME: &str = "stats-events-consumer";
+
+/// Subjects for events stream.
+pub const SUBJECTS: &str = "events.>";
```
tiagolobocastro, line 14: What does `>` signify?
Should subjects be configurable?

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,339 @@
...
+                STREAM_NAME,
+                err
+            );
+        }
+        Ok(())
+    }
+
+    /// Creates consumer and returns an iterator for the messages on the stream.
+    async fn create_consumer_and_get_messages(&mut self, stream: Stream) -> BusResult<Messages> {
+        tracing::info!("Getting/creating consumer for stats '{}'", CONSUMER_NAME);
```
tiagolobocastro: Since this is a library, should all of these info traces be debug?

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,339 @@
...
+                    "Error publishing message to jetstream: {:?}. Retrying...",
+                    err
+                );
+                let _stream = self.verify_stream_exists().await;
+                log_error = false;
+            }
+
+            if tries == options.max_retries {
+                let elapsed = now.elapsed();
+                tracing::info!("Elapsed: {:.2?}", elapsed);
```
tiagolobocastro: what does this log message mean?

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,339 @@
...
+        tracing::info!("Subscribed to Nats message bus successfully");
+
+        return Ok(BusSubscription {
+            messages,
+            _phantom: PhantomData::default(),
+        });
+    }
+}
+
+pub struct BusSubscription<T> {
```
tiagolobocastro, line 300: What does `T` represent here?

mbus-api/src/message.rs
```diff
@@ -0,0 +1,36 @@
+use serde::{Deserialize, Serialize};
```
tiagolobocastro, line 1: will all of these be changed to protobuf later?

mbus-api/src/common/constants.rs
datacore-vvarakantham: The wildcard ">" is for matching one or more tokens. Ex: subject for pool event is "events.pool.event_id" which is unique for that event. And with any number of publish retries the message will be published at most once, because the max_msgs_per_subject is set to 1 in the stream config while creation. [Infinite msg deduplication](https://nats.io/blog/new-per-subject-discard-policy/) is the link for reference. 

The other option is to set the duplicate window and retry the publish limited number of times(The duplicates will be checked only in the duplicate window time frame).

mbus-api/src/mbus_nats.rs
datacore-vvarakantham: The messages from bus are deserialised into the type T. T here is EventMessage since all the messages have same format.

mbus-api/src/message.rs
datacore-vvarakantham: If the messages in the buffer are protobuff, these might be changed.

mbus-api/src/common/constants.rs
datacore-vvarakantham: Changed it to 10 seconds.

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,337 @@
...
+        let mut tries = 0;
+        let mut log_error = true;
+
+        let subject = NatsMessageBus::subject(message);
+        let payload = NatsMessageBus::payload(message)?;
+
+        loop {
+            let publish_request = self.jetstream.publish(subject.clone(), payload.clone());
+            let err = match publish_request.await {
+                Ok(x) => match x.await {
```
tiagolobocastro: what is x? maybe rename to something a bit clearer

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,337 @@
...
+                    Ok(y) => {
```
tiagolobocastro: same here with y..

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,337 @@
...
+                },
+                Err(error) => error,
+            };
+
+            if log_error {
+                tracing::warn!(
+                    "Error publishing message to jetstream: {:?}. Retrying...",
+                    err
+                );
+                let _stream = self.verify_stream_exists().await;
```
tiagolobocastro: what does this line do?

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,337 @@
...
+            if tries == options.max_retries {
+                return Err(Error::PublishError {
+                    retries: options.max_retries,
+                    payload: format!("{message:?}"),
+                    error: err.to_string(),
+                });
+            }
+            if let Some(error) = err.downcast_ref::<std::io::Error>() {
+                if error.kind() == ErrorKind::TimedOut {
+                    tries += 1;
```
tiagolobocastro, line 272: shouldn't we also bump tries if it's a different error, otherwise we could get stuck looping forever?

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,337 @@
...
+            backoff_with_options(&mut tries, &options).await;
+        }
+    }
+
+    /// Subscribes to the stream on messages though a consumer. Returns an ordered stream of
+    /// messages published to the stream.
+    async fn subscribe<T: Serialize + DeserializeOwned>(
+        &mut self,
+    ) -> BusResult<BusSubscription<T>> {
+        tracing::debug!("Subscribing to Nats message bus");
```
tiagolobocastro: Tempted to say this should be trace rather than debug as we might have many subscriptions if we want to listen to various events?
And if so, should we also have a way to subscribe to many events using a single subscription, or is that not possible?

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,337 @@
...
+        return Ok(BusSubscription {
+            messages,
+            _phantom: PhantomData::default(),
+        });
+    }
+}
+
+/// The messages from bus are deserialised into the type T
+pub struct BusSubscription<T> {
+    pub messages: Messages,
```
tiagolobocastro: missing doc comment, also why does it need to be pub?

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,337 @@
...
+    _phantom: PhantomData<T>,
+}
+
+impl<T: Serialize + DeserializeOwned> BusSubscription<T> {
+    pub async fn next(&mut self) -> Option<T> {
```
tiagolobocastro, line 308: missing doc comment

mbus-api/src/mbus_nats.rs
```diff
@@ -0,0 +1,337 @@
...
+        loop {
+            if let Some(message) = self.messages.next().await {
+                let message = match message {
+                    Ok(message) => message,
+                    Err(error) => {
+                        tracing::warn!("Error accessing jetstream message: {:?}.", error);
```
tiagolobocastro: Consider adding fields on left side, %for display and ? for Debug.
```suggestion
                        tracing::warn!(%error, "Error accessing jetstream message");
```
Also trace messages should not end with period `.`

mbus-api/src/message.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub component: String,
+    /// Node name
+    pub node: String,
+}
+
+/// Event action.
+#[derive(Serialize, Deserialize, Debug)]
+pub enum Action {
+    CreateEvent,
+    DeleteEvent,
```
tiagolobocastro, line 42: Add Unkown variant so we can handle additions of new actions gracefully.

mbus-api/src/message.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/// Event category.
+#[derive(Serialize, Deserialize, Debug)]
+pub enum Category {
```
tiagolobocastro, line 48: Add Unkown variant so we can handle additions of new categories gracefully.

tiagolobocastro (approved):

mbus-api/src/mbus_nats.rs
datacore-vvarakantham: It checks if stream exists during first publish error and creates a stream if it doesn't exist.

mbus-api/src/mbus_nats.rs
datacore-vvarakantham: For any other error, the backoff_with_options function is incrementing the tries count.

mbus-api/src/mbus_nats.rs
tiagolobocastro: But we don't use _stream? It creates the stream inside nats? Then please add a comment because it's not obvious to a reader.

datacore-vvarakantham: A jetstream publish call requires just the subject. NATS will figure out the stream the message should be published to.

There might be a case that there is publish error because the stream is deleted for some reason. So this function checks and creates a stream if it doesn't exist. 

There might also be the case that the stream is deleted and so the consumer is also deleted on the subscription side. And there is a heartbeat mechanism on the subscription side for the consumer, to check if the consumer still exists at regular intervals and create if it doesn't exists. Which I didn't apply yet in the implementation.

mbus-api/src/mbus_nats.rs
datacore-vvarakantham: We can have different consumers to listen to different set of events. Here I was creating a single consumer to get all the events for now. I will change it to trace.

sinhaashish:
bors merge

PR #32, opened 2023-06-22
fix(util): fixing static memleak in version-info macro

dsavitskiy:


version-info/src/macros.rs
```diff
@@ -48,12 +48,17 @@ macro_rules! version_info_string {
...
 #[macro_export]
 macro_rules! version_info_str {
-    () => {
-        Box::leak(Box::new(String::from($crate::version_info!()))) as &'static str
-    };
+    () => {{
+        static mut BUF: Vec<u8> = Vec::new();
+        let s = String::from($crate::version_info!());
+        unsafe {
+            BUF.resize(s.len(), 0);
```
tiagolobocastro, line 57: Wouldn't a second call  to this invalidate the first memory as we're resizing?
would OnceCell be more appropriate usage?

tiagolobocastro:
My comment here was incorrect but anyhow I'm removing the macro instead since clap no longer requires a static string reference.

PR #33, opened 2023-06-26
chore(callhome): add protobuff event message

datacore-vvarakantham:
Proto message for event structure


mbus-api/protobuf/v1/event.proto
```diff
@@ -0,0 +1,44 @@
...
+
+// Event Category
+enum Category {
+  Unknown = 0;
+  Pool = 1;
+  Volume = 2;
+}
+
+// Event Action
+enum Action {
```
tiagolobocastro: Should be Event rather than Action>

mbus-api/protobuf/v1/event.proto
```diff
@@ -0,0 +1,44 @@
...
+  UnknownEvent = 0;
```
tiagolobocastro: Do these need the Event postfix?

mbus-api/protobuf/v1/event.proto
```diff
@@ -0,0 +1,44 @@
...
+  CreateEvent = 1;
+  DeleteEvent = 2;
+}
+
+// Event meta data
+message EventMeta {
+    // Something that uniquely identifies events. UUIDv4. GUID.
+    string id = 1;
+    EventSource source = 2;
+    string event_timestamp = 3;
```
tiagolobocastro:
```suggestion
    google.protobuf.Timestamp timestamp = 3;
```

mbus-api/protobuf/v1/event.proto
```diff
@@ -0,0 +1,44 @@
...
+
+import "google/protobuf/wrappers.proto";
+
+package v1.event;
+
+// Event Message
+message EventMessage {
+    Category category = 1;
+    Action action = 2;
+    string target = 3;
```
tiagolobocastro: what's the target?

mbus-api/protobuf/v1/event.proto
```diff
@@ -0,0 +1,44 @@
...
+    string id = 1;
+    EventSource source = 2;
+    string event_timestamp = 3;
+    string version = 4;
+}
+
+// Event source
+message EventSource {
+  // Io-engine or core-agent
+  string component = 1;
```
tiagolobocastro: should this be an enum as well?

mbus-api/src/mbus_nats.rs
```diff
@@ -212,13 +212,22 @@ impl NatsMessageBus {
...
+    fn subject(msg: &EventMessage) -> BusResult<String> {
+        let event_id = match &msg.metadata {
+            Some(event_meta) => &event_meta.id,
+            None => {
+                return Err(Error::InvalidMessageId {
+                    error_msg: "the message id must not be empty".to_string(),
+                })
+            }
+        };
+        let subject = format!("events.{}.{}", msg.category, event_id); // If category is volume
```
tiagolobocastro: comment is hard to read, add it on top rather than short multi-line

mbus-api/protobuf/v1/event.proto
```diff
@@ -0,0 +1,44 @@
...
+  DeleteEvent = 2;
+}
+
+// Event meta data
+message EventMeta {
+    // Something that uniquely identifies events. UUIDv4. GUID.
+    string id = 1;
+    EventSource source = 2;
+    string event_timestamp = 3;
+    string version = 4;
```
tiagolobocastro: should this be an enum?

mbus-api/protobuf/v1/event.proto
datacore-vvarakantham: Changed it to EventAction.

mbus-api/protobuf/v1/event.proto
datacore-vvarakantham: Removed the postfix.

mbus-api/protobuf/v1/event.proto
datacore-vvarakantham: Id for the category against which action is performed. Added comment.

mbus-api/protobuf/v1/event.proto
datacore-vvarakantham: Currently I am getting the component from the user.

mbus-api/protobuf/v1/event.proto
datacore-vvarakantham: What could be the enum variants? UnknownVersion, v1? Currently I am just using "v1".

sinhaashish:
Rename this file `event_traits.rs` to `events` .IMHO appending traits to file name is not idomatic.

sinhaashish (commented):
lgtm

sinhaashish (approved):

mbus-api/Cargo.toml
```diff
@@ -14,3 +18,8 @@ uuid = { version = "1.3.0", features = ["v4"] }
 async-trait = "0.1.64"
 futures = "0.3.26"
 snafu = "0.7.4"
+tonic = "0.8.3"
+prost = "0.11.6"
+prost-wkt-types = "0.4.0"
```
tiagolobocastro, line 23: do we need wkt? We were able to do without this in control-plane at the cost of adding our own conversions but avoiding pulling in the extra code

mbus-api/build.rs
```diff
@@ -0,0 +1,7 @@
+fn main() {
+    tonic_build::configure()
+        .type_attribute(".", "#[derive(serde::Deserialize, serde::Serialize)]")
+        .extern_path(".google.protobuf.Timestamp", "::prost_wkt_types::Timestamp")
```
tiagolobocastro: same here

mbus-api/protobuf/v1/event.proto
```diff
@@ -0,0 +1,65 @@
...
+    // Event action
+    EventAction action = 2;
+    // Target id for the category against which action is performed
+    string target = 3;
+    // Event meta data
+    EventMeta metadata = 4;
+}
+
+// Event Category
+enum EventCategory {
```
tiagolobocastro: Ah so here is Category, so perhaps my suggestion of Resource is not correct @sinhaashish ..

mbus-api/src/event_traits.rs
```diff
@@ -0,0 +1,55 @@
+use crate::event::{Component, EventMessage, EventMeta, EventSource, Version};
+use chrono::Utc;
+use once_cell::sync::OnceCell;
+use std::str::FromStr;
+
+/// Once cell static variable to store the component field of the event source.
+static COMPONENT: OnceCell<Component> = OnceCell::new();
+
+/// Initialize the event source component with the service name.
+pub fn initilize_source_component(comp: String) {
```
tiagolobocastro: This will be more useful as &String can be coerced into &str
```suggestion
pub fn initilize_source_component(comp: &str) {
```

mbus-api/src/event_traits.rs
```diff
@@ -0,0 +1,55 @@
...
+            timestamp: Some(Utc::now().into()),
+            version: Version::V1.into(),
+        }
+    }
+}
+
+impl EventSource {
+    /// New event source with default values.
+    pub fn new() -> Self {
+        let component = (*COMPONENT.get().unwrap_or(&Component::UnknownComponent)).into();
```
tiagolobocastro: Can we use unwrap_or_default here too?

mbus-api/src/event_traits.rs
```diff
@@ -0,0 +1,55 @@
...
+        Self {
+            component,
+            node: "".to_string(),
```
tiagolobocastro: Should set the node here?

mbus-api/src/event_traits.rs
```diff
@@ -0,0 +1,55 @@
...
+        tracing::event!(target: "target-mbus", tracing::Level::INFO, event_data);
+    }
+}
+
+// Get Component from service name.
+impl FromStr for Component {
+    type Err = String;
+    fn from_str(c: &str) -> Result<Self, Self::Err> {
+        match c {
+            "core-agent" => Ok(Self::CoreAgent),
```
tiagolobocastro: the app label is agent-core so I suggest using that instead?

mbus-api/Cargo.toml
datacore-vvarakantham: prost_wkt_types is used in the control plane when I last checked. Can you please give an example for the conversion if not?

mbus-api/src/event_traits.rs
datacore-vvarakantham: I am using the default value of null string here. And setting the node value from the data in event trait. Should I use a param and set the node value here?

tiagolobocastro: hmm where is it used, maybe new should be called there with node as parameter?

mbus-api/protobuf/v1/event.proto
sinhaashish: Let @datacore-vvarakantham  PR get in, will update my changes accordingly , There are few renames done in her PR

mbus-api/Cargo.toml
tiagolobocastro: It is not used in the control-plane for sure... please point me to where it is otherwise

mbus-api/protobuf/v1/event.proto
```diff
@@ -0,0 +1,65 @@
...
+  UnknownCategory = 0;
+  Pool = 1;
+  Volume = 2;
+}
+
+// Event Action
+enum EventAction {
+  UnknownAction = 0;
```
tiagolobocastro: nit: Don't need the Action postfix, it's already implicit in the type

mbus-api/protobuf/v1/event.proto
```diff
@@ -0,0 +1,65 @@
...
+    EventAction action = 2;
+    // Target id for the category against which action is performed
+    string target = 3;
+    // Event meta data
+    EventMeta metadata = 4;
+}
+
+// Event Category
+enum EventCategory {
+  UnknownCategory = 0;
```
tiagolobocastro: nit: Don't need the Category postfix, it's already implicit in the type

tiagolobocastro (approved):

mbus-api/protobuf/v1/event.proto
datacore-vvarakantham: I was facing error while adding 2 enum variants of same type(unknown) in one module. So I used the postfix.

mbus-api/Cargo.toml
datacore-vvarakantham: Will work on adding the serde-proto crate with our own timestamp in the next PR.

PR #35, opened 2023-07-24
feat(callhome): change the events tracing target from 'mbus-target' to 'mbus-events-target'

datacore-vvarakantham:


events-api/src/event_traits.rs
```diff
@@ -35,7 +35,7 @@ impl EventMessage {
     /// Generate the event trace with event message.
     pub fn generate(&self) {
         let event_data = serde_json::to_string(&self).unwrap_or_default();
-        tracing::event!(target: "target-mbus", tracing::Level::INFO, event_data);
+        tracing::event!(target: "mbus-events-target", tracing::Level::INFO, event_data);
```
tiagolobocastro, line 38: btw, should we only add event in case serde_json above is ok? Atm we're doing it for default which will be bogus if json to string fails..

tiagolobocastro (approved):

events-api/src/event_traits.rs
datacore-vvarakantham: What should be done if there is an error? Should I print the error message to debug?

tiagolobocastro: Yeah perhaps? Though might be better to leave  as is, and let event subscribers pick up, presumably they'd flag receiving empty String when they expect an event type?

datacore-vvarakantham: Yes. Stats module will probably trace the unknown events. Will have to check.

datacore-vvarakantham: The message will not be published to the mbus, as the message id is empty. The events-api lib gives the InvalidMessageId error.

sinhaashish (approved):

PR #36, opened 2023-08-07
feat(eventing): init with optional spawn function

datacore-vvarakantham:
With this code, we can use task spawning functionality in data plane

### DIFF
--- a/event-publisher/src/event_handler.rs
+++ b/event-publisher/src/event_handler.rs
@@ -9,14 +9,26 @@ pub struct EventHandle {}
 impl EventHandle {
     /// Initilize the buffer, starts the event publisher and return the layer for handling the event
     /// traces.
-    pub fn init(mbus_url: String, service_name: &str) -> EventLayer {
+    pub fn init<T>(mbus_url: String, service_name: &str, spawn_option: Option<T>) -> EventLayer
+    where
+        T: Fn(std::pin::Pin<Box<dyn core::future::Future<Output = ()> + Send>>),
+    {
         let (send, recv) = tokio::sync::mpsc::channel::<EventMessage>(MAX_BUFFER_MSGS);
         initilize_source_component(service_name);
         let layer = EventLayer::new(send);
-        tokio::spawn(async move {
+        let publisher_future = Box::pin(async move {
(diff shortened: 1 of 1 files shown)


event-publisher/src/event_handler.rs
```diff
@@ -1,22 +1,37 @@
...
         initilize_source_component(service_name);
         let layer = EventLayer::new(send);
-        tokio::spawn(async move {
+        let publisher_future = Box::pin(async move {
             MbusPublisher::run(&mbus_url, recv).await;
-            // TODO handle the closed channel situation.
         });
+        match spawn_option {
+            Some(spawn_fn) => {
+                // Spawn the publisher on the tokio-runtime within mayastor.
```
tiagolobocastro:
```suggestion
                // Spawn the publisher on the spawner specified in the args.
```

event-publisher/src/event_handler.rs
```diff
@@ -1,22 +1,37 @@
...
+                spawn_fn(publisher_future);
+            }
+            None => {
+                tokio::spawn(publisher_future);
+            } // TODO handle the closed channel situation.
```
tiagolobocastro: Should this comment be here on the the publisher_future above?

event-publisher/src/event_handler.rs
```diff
@@ -1,22 +1,37 @@
...
 
 /// Event handle.
-pub struct EventHandle {}
+pub struct EventHandle<T> {
+    __phantom: PhantomData<T>,
+}
 
-impl EventHandle {
+impl<T> EventHandle<T>
+where
```
tiagolobocastro: You don't need to add T for the struct btw, I think can just add it to the init function

tiagolobocastro (approved):

tiagolobocastro (commented):
rename PR and commit to simply `feat(eventing): init with optional spawn function`

event-publisher/src/event_handler.rs
datacore-vvarakantham: No. It should be for publisher_future. Will move it.

sinhaashish (approved):
LGTM

PR #37, opened 2023-08-08
feat(eventing): add nexus events in protobuf

datacore-vvarakantham:
Added nexus category and rebuild, stateChange actions in event protobuf

### DIFF
--- a/events-api/protobuf/v1/event.proto
+++ b/events-api/protobuf/v1/event.proto
@@ -23,6 +23,7 @@ enum EventCategory {
   UnknownCategory = 0;
   Pool = 1;
   Volume = 2;
+  Nexus = 3;
 }
 
 // Event Action
@@ -30,6 +31,9 @@ enum EventAction {
   UnknownAction = 0;
   Create = 1;
   Delete = 2;
+  StateChange = 3;
+  RebuildBegin = 4;
+  RebuildEnd = 5;
 }
 
 // Event meta data


events-api/protobuf/v1/event.proto
```diff
@@ -23,13 +23,17 @@ enum EventCategory {
...
+  Nexus = 3;
 }
 
 // Event Action
 enum EventAction {
   UnknownAction = 0;
   Create = 1;
   Delete = 2;
+  StateChange = 3;
+  RebuildBegin = 4;
```
tiagolobocastro, line 35: Are we simply interested in begin/end, or do we want to specify rebuild source/destination? If so, where would event specific data go into?

datacore-vvarakantham: We can add data specific to particular event categories.

sinhaashish: But why do the stats need source and destination. IMO source and destination of rebuilds are for logs, where we need to debug. But why to show that in reports?

tiagolobocastro: Stats? I thought these were generic events and stats is just one use-case ?

sinhaashish: hmm, for other use case it make sense. i though wrt to stats only.

datacore-vvarakantham: We need source/destination for rebuilds. I will work on the event specific data in the future PRs.

chriswldenyer (commented):
Looks OK.
Ideally we would like event-specific data fo the rebuild, but that can be in a later PR.

chriswldenyer (approved):

tiagolobocastro (approved):

PR #40, opened 2023-09-06
feat(eventing): add event specific data

datacore-vvarakantham:
Adding event specific data, rebuild events specific data.

Adding RebuildStatus(Started, Completed, Stopped, Failed) for rebuild events.

### DIFF
--- a/events-api/protobuf/v1/event.proto
+++ b/events-api/protobuf/v1/event.proto
@@ -53,6 +53,8 @@ message EventSource {
   Component component = 1;
   // Node name
   string node = 2;
+  // Event specific details
+  optional EventDetails event_details = 3;
 }
 
 // Source Component
@@ -66,4 +68,36 @@ enum Component {
 enum Version {
   UnknownVersion = 0;
   V1 = 1;
+}
+
+// Event specific details
+message EventDetails {
+  // Rebuild event details
+  optional RebuildEventDetails rebuild_details = 1;
+}
+
+// Rebuild event details
+message RebuildEventDetails {
+  // Rebuild status
+  RebuildStatus rebuild_status = 1;
+  // Rebuild source replica uri
+  string source_replica = 2;
+  // Rebuild destination replica uri
+  string destination_replica = 3;
+  // Rebuild error for RebuildFail event
+  optional string error = 4;
+}
+
+// Rebuild status
+enum RebuildStatus {
+  // Unknown
+  Unknown = 0;
+  // Rebuild started
+  Started = 1;
+  // Rebuild is completed successfully
+  Completed = 2;
+  // Rebuild is stopped
+  Stopped = 3;
+  // Rebuild failed
+  Failed = 4;
 }
--- a/events-api/src/event_traits.rs
+++ b/events-api/src/event_traits.rs
@@ -1,4 +1,7 @@
-use crate::event::{Component, EventMessage, EventMeta, EventSource, Version};
+use crate::event::{
+    Component, EventDetails, EventMessage, EventMeta, EventSource, RebuildEventDetails,
+    RebuildStatus, Version,
+};
 use chrono::Utc;
 use once_cell::sync::OnceCell;
 use std::str::FromStr;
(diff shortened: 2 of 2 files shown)


sinhaashish (approved):

events-api/protobuf/v1/event.proto
```diff
@@ -34,6 +34,8 @@ enum EventAction {
   StateChange = 3;
   RebuildBegin = 4;
   RebuildEnd = 5;
+  RebuildStop = 6;
+  RebuildFail = 7;
```
tiagolobocastro: Aren't these just RebuildEnd with failure? Perhaps we can't use metadata to signify this instead of adding new events?

datacore-vvarakantham: RebuildEnd here is generated when rebuild is completed. Should I instead change the RebuildEnd to RebuildComplete? Or should I use single RebuildEnd (with stop, fail and complete in the metadata)?

tiagolobocastro: Yeah that's what I'm wondering.. I'm leaning more towards the latter one, end with bool flag in the event specific data. @chriswldenyer wdyt?

chriswldenyer: Yes I think so as well. There could be cases where we want to know if a rebuild has completed but not care about success or failure. If we do care, we look at the event-specific data.

datacore-vvarakantham: Done

events-api/protobuf/v1/event.proto
```diff
@@ -67,3 +69,33 @@ enum Version {
...
+// Event specific details
+message EventDetails {
+  // Rebuild event details
+  optional RebuildEventDetails rebuild_details = 1;
+}
+
+// Rebuild event details
+message RebuildEventDetails {
+  // Rebuild end status
+  optional RebuildEndStatus rebuild_end_status = 1;
```
tiagolobocastro: Is this optional for the case of start rebuild?
In that case I suggest renaming RebuildEndStatus to RebuildStatus and adding a Started there.

datacore-vvarakantham: Yes. It was optional for start rebuild. Will change it.

events-api/src/event_traits.rs
```diff
@@ -35,7 +38,32 @@ impl EventSource {
...
+            node,
+            ..Default::default()
+        }
+    }
+
+    /// Add rebuild event specific data to event source.
+    pub fn add_rebuild_data(
+        self,
+        status: RebuildStatus, // Rebuild status
+        source: Option<&str>,  // Rebuild source replica uri
```
tiagolobocastro: why is source optional?

events-api/protobuf/v1/event.proto
```diff
@@ -67,3 +69,35 @@ enum Version {
...
+// Rebuild event details
+message RebuildEventDetails {
+  // Rebuild status
+  RebuildStatus rebuild_status = 1;
+  // Rebuild source replica uri
+  string source_replica = 2;
+  // Rebuild destination replica uri
+  string destination_replica = 3;
+  // Rebuild error for RebuildFail event
+  string error = 4;
```
tiagolobocastro: error should be optional?

tiagolobocastro (approved):

events-api/src/event_traits.rs
datacore-vvarakantham: Made it optional for the rebuild events where I couldn't find the source details.

events-api/protobuf/v1/event.proto
datacore-vvarakantham: Yes. Will change it. Was using empty string when there is no error.

events-api/src/event_traits.rs
tiagolobocastro: Could you point me to such a place?

datacore-vvarakantham: Removed optional as the source_uri can be found for all the rebuild events.

PR #41, opened 2023-09-25
feat(eventing): add replica, node, HA event categories to protobuff

datacore-vvarakantham:
Added replica, node, HA(high availability) event categories.
Added replica event specific details(pool name).

### DIFF
--- a/events-api/protobuf/v1/event.proto
+++ b/events-api/protobuf/v1/event.proto
@@ -24,6 +24,9 @@ enum EventCategory {
   Pool = 1;
   Volume = 2;
   Nexus = 3;
+  Replica = 4;
+  Node = 5;
+  HighAvailability = 6;
 }
 
 // Event Action
@@ -74,6 +77,8 @@ enum Version {
 message EventDetails {
   // Rebuild event details
   optional RebuildEventDetails rebuild_details = 1;
+  // Replica event details
+  optional ReplicaEventDetails replica_details = 2;
 }
 
 // Rebuild event details
@@ -100,4 +105,14 @@ enum RebuildStatus {
   Stopped = 3;
   // Rebuild failed
   Failed = 4;
+}
+
+// Replica event details
+message ReplicaEventDetails {
+  // Pool name
+  string pool_name = 1;
+  // Pool uuid
+  string pool_uuid = 2;
+  // Replica name
+  string replica_name = 3;
 }
--- a/events-api/src/event_traits.rs
+++ b/events-api/src/event_traits.rs
@@ -1,6 +1,6 @@
 use crate::event::{
     Component, EventDetails, EventMessage, EventMeta, EventSource, RebuildEventDetails,
-    RebuildStatus, Version,
+    RebuildStatus, ReplicaEventDetails, Version,
 };
 use chrono::Utc;
 use once_cell::sync::OnceCell;
@@ -46,7 +46,7 @@ impl EventSource {
     }
 
     /// Add rebuild event specific data to event source.
-    pub fn add_rebuild_data(
+    pub fn with_rebuild_data(
         self,
         status: RebuildStatus, // Rebuild status
         source: &str,          // Rebuild source replica uri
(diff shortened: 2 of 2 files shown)


events-api/protobuf/v1/event.proto
```diff
@@ -24,6 +24,9 @@ enum EventCategory {
   Pool = 1;
   Volume = 2;
   Nexus = 3;
+  Replica = 4;
+  Node = 5;
+  HA = 6;
```
tiagolobocastro: what is this HA?

events-api/src/event_traits.rs
```diff
@@ -61,6 +61,23 @@ impl EventSource {
...
+                ..Default::default()
+            }),
+            ..self
+        }
+    }
+
+    /// Add replica event specific data to event source.
+    pub fn add_replica_data(
+        self,
+        pool_name: &str, // Pool name
```
tiagolobocastro:
```suggestion
        pool_name: &str,
```

events-api/src/event_traits.rs
```diff
@@ -61,6 +61,23 @@ impl EventSource {
...
                     error,
                 }),
+                ..Default::default()
+            }),
+            ..self
+        }
+    }
+
+    /// Add replica event specific data to event source.
+    pub fn add_replica_data(
```
tiagolobocastro: Did you mean to move self?
If so then suggest with_ prefix to make it clearer.

```suggestion
    pub fn with_replica_data(
```

events-api/protobuf/v1/event.proto
datacore-vvarakantham: It is for high availability events. Currently a single event action "SwitchOver" is in the list of events in this category.

events-api/src/event_traits.rs
datacore-vvarakantham: Yes I am moving self. Changed it now.

events-api/protobuf/v1/event.proto
```diff
@@ -101,3 +106,9 @@ enum RebuildStatus {
   // Rebuild failed
   Failed = 4;
 }
+
+// Replica event details
+message ReplicaEventDetails {
+  // Pool name
+  string pool = 1;
```
tiagolobocastro: What about the replica uuid?

tiagolobocastro: Btw let's add optional pool_uuid as well

datacore-vvarakantham: Replica uuid is the "lvol.name()" which is added to target. I can add "lvol.uuid()" to the event details.

sinhaashish (approved):

events-api/protobuf/v1/event.proto
datacore-vvarakantham: Should the pool_uuid be optional?

tiagolobocastro: Ah this is coming from dataplane now, so should not be of "optional" kind.
Replica uuid being name is just current way control-plane creates replicas, this may change in the future.

events-api/protobuf/v1/event.proto
datacore-vvarakantham: Changed it to "HighAvailability"

events-api/protobuf/v1/event.proto
datacore-vvarakantham: Done

events-api/protobuf/v1/event.proto
```diff
@@ -101,3 +106,13 @@ enum RebuildStatus {
...
 }
+
+// Replica event details
+message ReplicaEventDetails {
+  // Pool name
+  string pool_name = 1;
+  // Pool uuid
+  string pool_uuid = 2;
+  // lvol uuid
+  string lvol_uuid = 3;
```
tiagolobocastro: Lvol is impl specific

```suggestion
  string replica_uuid = 3;
```

Btw should it be the other way around, uuid in the target and name here?

datacore-vvarakantham: Yes the uuid should be in the target. 

I was looking at from<lvol> for replica in a different location. Will change it.

datacore-vvarakantham: Done.

tiagolobocastro (approved):

PR #43, opened 2023-10-11
feat(eventing): add proto serializable timestamp

datacore-vvarakantham:
Added serve-proto crate for serialisable timestamp. Remove dependency on prost-wkt-types::Timestamp.

### DIFF
--- a/events-api/Cargo.toml
+++ b/events-api/Cargo.toml
@@ -23,3 +23,4 @@ prost = "0.11.9"
 prost-wkt-types = "0.4.2"
 chrono = "0.4.31"
 once_cell = "1.18.0"
+prost-extend = { path = "../prost-extend" }
--- a/events-api/build.rs
+++ b/events-api/build.rs
@@ -1,7 +1,7 @@
 fn main() {
     tonic_build::configure()
         .type_attribute(".", "#[derive(serde::Deserialize, serde::Serialize)]")
-        .extern_path(".google.protobuf.Timestamp", "::prost_wkt_types::Timestamp")
+        .extern_path(".google.protobuf.Timestamp", "::prost_extend::Timestamp")
         .compile(&["protobuf/v1/event.proto"], &["protobuf/"])
         .unwrap_or_else(|e| panic!("event v1 protobuf compilation failed: {e}"));
 }
--- a/events-api/src/event_traits.rs
+++ b/events-api/src/event_traits.rs
@@ -124,7 +124,7 @@ mod test {
         let event_source = EventSource::new("".to_string());
         let event_meta = EventMeta::from_source(event_source);
         assert!(!event_meta.id.is_empty());
-        assert!(!event_meta.timestamp.unwrap().to_string().is_empty());
+        assert!(event_meta.timestamp.is_some());
         assert_eq!(event_meta.version, Version::V1 as i32);
         assert_ne!(event_meta.source, None);
     }
--- a/prost-extend/Cargo.toml
+++ b/prost-extend/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "prost-extend"
+version = "0.1.0"
+edition = "2021"
+
+
+[build-dependencies]
+tonic-build = "0.8.4"
+
+[dependencies]
+chrono = "0.4.26"
+serde = "1.0.183"
+tonic = "0.8.3"
+prost = "0.11.9"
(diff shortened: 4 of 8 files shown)


serde-proto/src/timestamp.rs
```diff
@@ -0,0 +1,69 @@
...
+use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
+
+use crate::common_types::Timestamp;
+
+impl TryFrom<Timestamp> for DateTime<Utc> {
+    type Error = &'static str;
+
+    fn try_from(value: Timestamp) -> Result<Self, Self::Error> {
+        let dt = DateTime::<Utc>::from_timestamp(value.seconds, value.nanos)
+            .expect("The datetime is not valid or falls outside the valid range");
```
tiagolobocastro: This should return error rather than expect?

serde-proto/protobuf/v1/common_types.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+import "google/protobuf/wrappers.proto";
+
+package v1.common_types;
+
+
+message Timestamp {
```
tiagolobocastro: We don't need to define our own, we just add
import "google/protobuf/timestamp.proto";
and then use this `Timestamp` ?

serde-proto/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "serde-proto"
```
tiagolobocastro: it's not just serde but also date-time so maybe suggest renaming to something else?
Not a good suggestion, but example could do similar to the wtk crate:
wkt-proto

serde-proto/protobuf/v1/common_types.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+import "google/protobuf/wrappers.proto";
+
+package v1.common_types;
```
tiagolobocastro: Separate common into specific protos if we need more kinds in the future
```suggestion
package v1.pb_time;
```

serde-proto/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+mod timestamp;
+
+/// Event module for the autogenerated event code.
+pub mod common_types {
+    #![allow(clippy::derive_partial_eq_without_eq)]
+    #![allow(clippy::large_enum_variant)]
+    tonic::include_proto!("v1.common_types");
+}
```
tiagolobocastro: Control what we export?
```suggestion
}
pub use common_types::Timestamp;
```

serde-proto/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+mod timestamp;
+
+/// Event module for the autogenerated event code.
+pub mod common_types {
```
tiagolobocastro:
```suggestion
mod common_types {
```

serde-proto/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
+edition = "2021"
+
+
+[build-dependencies]
+tonic-build = "0.8.4"
+
+[dependencies]
+tracing = "0.1.37"
```
tiagolobocastro: Is tracing an tonic/tonic-build needed here?

serde-proto/protobuf/v1/common_types.proto
datacore-vvarakantham: Yes. We need to also add
extern_path(".google.protobuf.Timestamp", "::serde_extend::Timestamp")
 while build the protobuf to use this timestamp

serde-proto/Cargo.toml
datacore-vvarakantham: Do  I use 
prost-utils
prost-helpers
Prost-common-utils

Similar to these?

tiagolobocastro: Not keen on the last one but the others sound better.
I'll add a few more to choose from
prost-boost
prost-extend

serde-proto/src/timestamp.rs
datacore-vvarakantham: Yes, returning the error now.

serde-proto/Cargo.toml
datacore-vvarakantham: Using prost-extend.

serde-proto/protobuf/v1/common_types.proto
datacore-vvarakantham: Done.

serde-proto/Cargo.toml
datacore-vvarakantham: Tracing is not needed. Removed it now. The others are needed.

tiagolobocastro (approved):

chriswldenyer (approved):
Looks believable.

PR #49, opened 2023-11-08
fix(nvmeadm/addr): match for host&port only

tiagolobocastro:
When listing we can't compare the full address with a host:port because the address from the subsystem might have other components such as the source address.
Instead, let's only check if this is the host:port we're interested in.

### DIFF
--- a/nvmeadm/src/nvmf_subsystem.rs
+++ b/nvmeadm/src/nvmf_subsystem.rs
@@ -5,13 +5,13 @@ use error::{
 };
 use glob::glob;
 use snafu::ResultExt;
-use std::{fs::OpenOptions, io::Write, path::Path, str::FromStr};
+use std::{collections::HashMap, fs::OpenOptions, io::Write, path::Path, str::FromStr};
 
 pub const SYSFS_NVME_CTRLR_PREFIX: &str = "/sys/devices/virtual/nvme-fabrics/ctl";
 
 /// Subsystem struct shows us all the connect fabrics. This does not include
 /// NVMe devices that are connected by trtype=PCIe.
-#[derive(Default, Clone, Debug)]
+#[derive(Clone, Debug)]
 pub struct Subsystem {
     /// Name of the subsystem.
     pub name: String,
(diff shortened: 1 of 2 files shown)


nvmeadm/src/nvmf_subsystem.rs
```diff
@@ -23,26 +23,114 @@ pub struct Subsystem {
...
     pub model: String,
 }
 
-/// Wrapper structure that creates subsystem addr string.
+const TR_ADDR: &str = "traddr";
+const TR_SVC_ID: &str = "trsvcid";
+const SRC_ADDR: &str = "src_addr";
+
+#[derive(Debug, Clone)]
+#[allow(unused)]
```
Abhinandan-Purkait, line 40: Is this needed?

Abhinandan-Purkait (approved):

hrudaya21 (approved):

nvmeadm/src/nvmf_subsystem.rs
tiagolobocastro: Yes, because we're not using unknown atm

nvmeadm/src/nvmf_subsystem.rs
```diff
@@ -23,26 +23,114 @@ pub struct Subsystem {
...
 impl SubsystemAddr {
-    /// New SubsystemAddr.
-    pub fn new(host: String, port: u16) -> SubsystemAddr {
-        SubsystemAddr(format!("traddr={host},trsvcid={port}"))
+    /// Check if the subsystem address contains target port.
+    pub fn match_host_port(&self, host: &str, port: &str) -> bool {
+        SubsystemAddrExt::from(self.as_str()).match_host_port(host, port)
+    }
+    /// Collect all address variants into a key-val map.
+    pub fn collect(&self) -> HashMap<&str, &str> {
```
datacore-vvarakantham, line 120: Is the collect function used?

datacore-vvarakantham (approved):

nvmeadm/src/nvmf_subsystem.rs
tiagolobocastro: No, left it as part of pub in case we need to inspect the key-map

PR #50, opened 2023-11-23
feat(eventing): add nexus online child event

datacore-vvarakantham:
Added OnlineChild action to events proto.

### DIFF
--- a/apis/events/protobuf/v1/event.proto
+++ b/apis/events/protobuf/v1/event.proto
@@ -44,6 +44,7 @@ enum EventAction {
   NvmePathSuspect = 9;
   NvmePathFail = 10;
   NvmePathFix = 11;
+  OnlineChild = 12;
 }
 
 // Event meta data


apis/events/protobuf/v1/event.proto
```diff
@@ -41,9 +41,10 @@ enum EventAction {
   SwitchOver = 6;
   AddChild = 7;
   RemoveChild = 8;
-  NvmePathSuspect = 9;
-  NvmePathFail = 10;
-  NvmePathFix = 11;
+  OnlineChild = 9;
```
tiagolobocastro: Might be worth clarifying if this is RPC of childState change, maybe like so:
```suggestion
  AddChildRpc = 7;
  RemoveChildRpc = 8;
  OnlineChildRpc = 9;
```
Wdyt?
Not sure why this is showing suggestion removing nvme path ones, not intended...

tiagolobocastro: I take it back, as per name `OnlineChild` suggest it's an action, rather than `ChildOnline` or `ChildIsOnline`.

tiagolobocastro (approved):

apis/events/protobuf/v1/event.proto
datacore-vvarakantham: Ok. Yes that would be better.

Actually the add child, remove child, online child events are only being generated if new replicas are added to a nexus according to the current events generation. Should this be okay?

chriswldenyer (approved):
Looks OK but I am slightly bothered by moving existing values up when we insert new enums. This would break existing code which expects e.g. 9 == NVMEPathSuspect.
Would it be worth doing:
...
OnlineChild = 10;
NvmePathSuspect = 100;
NvmePathFail = 101;
...
SomeOtherKindCreate = 200;
etc?

datacore-vvarakantham:
@chriswldenyer the events are not yet merged in develop from control plane. So I think this change will not be a problem?

chriswldenyer:
> @chriswldenyer the events are not yet merged in develop from control plane. So I think this change will not be a problem?

True, but I do have e2e-test code which translates those number into strings (I've been "left-shifting"), so I will need to do a PR.
In the future we might want to insert new events which would shift existing published enums.

datacore-vvarakantham:
Ok. Yes we should not change the values. We should either add the new values after the existing ones or should follow the the method similar to the one you mentioned to add the values.

PR #53, opened 2023-11-28
feat(eventing): add host events

datacore-vvarakantham:
Added host initiatorCconnect, Disconnect and KeepAliveTimeout events.

### DIFF
--- a/apis/events/protobuf/v1/event.proto
+++ b/apis/events/protobuf/v1/event.proto
@@ -28,6 +28,7 @@ enum EventCategory {
   Node = 5;
   HighAvailability = 6;
   NvmePath = 7;
+  HostInitiator = 8;
 }
 
 // Event Action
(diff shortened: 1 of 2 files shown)


apis/events/protobuf/v1/event.proto
```diff
@@ -45,6 +46,9 @@ enum EventAction {
   NvmePathFail = 10;
   NvmePathFix = 11;
   OnlineChild = 12;
+  Connect = 13;
+  Disconnect = 14;
+  KeepAliveTimeout = 15;
```
tiagolobocastro: qualify it? CC @chriswldenyer 

```suggestion
  NvmeConnect = 13;
  NvmeDisconnect = 14;
  NvmeKeepAliveTimeout = 15;
```

tiagolobocastro (approved):

apis/events/protobuf/v1/event.proto
chriswldenyer: Yes, I would agree with that.

chriswldenyer (approved):
Looks reasonable, but with the nvme prefix please.

PR #55, opened 2023-12-07
feat(io-engine/api/cksum): add wipe checksum

tiagolobocastro:
Checksums a replica which can be used to verify data between volume replicas. For speed sake we're reusing the existing wipe replicas as it's a very good fit for this and we can reuse the entire infra around it. todo: consider renaming wipe etc


hrudaya21 (approved):

blaisedias:
Can we return the capabilities in mayastor features perhaps under test flags by extending the proto to add test capability flags to
```
message MayastorFeatures {
  bool asymmetricNamespaceAccess = 1;
}
```
in host.proto

Abhinandan-Purkait (approved):

PR #57, opened 2024-01-05
ci: add script to fetch git org name

tiagolobocastro:
This can be used by other repos to figure out the name of the org.

### DIFF
--- a/scripts/git-org-name.sh
+++ b/scripts/git-org-name.sh
@@ -1,0 +1,114 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+die() {
+  local _return="${2:-1}"
+  echo "$1" >&2
+  exit "${_return}"
+}
+
+# Returns the git orgname from a given git url (both https and git)
+# Example: git@github.com:openebs/mayastor.git => openebs
+urlOrgName() {
+  local gitUrl="$1"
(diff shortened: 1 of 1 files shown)


pchandra19 (approved):

scripts/git-org-name.sh
```diff
@@ -0,0 +1,100 @@
...
+    echo "$gitUrl" | awk -F/ '/git/ { print $1 }' | cut -d':' -f2
+  elif echo $gitUrl | grep -qa "^https://"; then
+    echo "$gitUrl" | awk -F/ '/https/ { print $4 }'
+  else
+    die "Unknown git url type: '$gitUrl'"
+  fi
+}
+
+orgName() {
+  local originUrl=$(git config --get remote.origin.url)
```
niladrih: Is the remote always guaranteed to be 'origin'? Is this meant for automation only? I call the openebs org 'upstream' in my dev setup.

niladrih (approved):
Left one comment, but LGTM.

scripts/git-org-name.sh
pchandra19: Since we clone , I would say yes.

tiagolobocastro: It's meant for the CI so we can name dockerhub credentials and also dockerhub repository name.
We can add a manual repository override for local image building if required.

PR #58, opened 2024-01-08
feat(eventing): add io-engine events

datacore-vvarakantham:
Added io-engine start, stop, shutdown events.
Reactor freeze, unfreeze events.
Nexus shutdown event.

### DIFF
--- a/apis/events/protobuf/v1/event.proto
+++ b/apis/events/protobuf/v1/event.proto
@@ -29,6 +29,7 @@ enum EventCategory {
   HighAvailability = 6;
   NvmePath = 7;
   HostInitiator = 8;
+  IoEngineCategory = 9;
 }
 
 // Event Action
@@ -49,6 +50,11 @@ enum EventAction {
   NvmeConnect = 13;
   NvmeDisconnect = 14;
   NvmeKeepAliveTimeout = 15;
+  ReactorFreeze = 16;
+  ReactorUnfreeze = 17;
+  Shutdown = 18;
+  Start = 19;
+  Stop = 20;
 }
 
 // Event meta data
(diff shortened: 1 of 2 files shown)


apis/events/protobuf/v1/event.proto
```diff
@@ -186,3 +195,17 @@ message HostInitiatorEventDetails {
...
+// Reactor event details
+message ReactorEventDetails {
+  // The logical core this reactor is created on
+  uint64 lcore = 1;
+  // Reactor state
+  string state = 2;
+}
+
+// Nexus shutdown event details
+message NexusShutdownEventDetails {
```
tiagolobocastro: Shouldn't it be IoEngine Shutdown? (it's the whole process being shutdown, not just the nexuses)
Then by its side we can have a per-nexus shutdown event?

datacore-vvarakantham: Should we have two separate events for io-engine stop and shutdown?

datacore-vvarakantham: Added IoEngine shutdown and per nexus shutdown events.

apis/events/protobuf/v1/event.proto
```diff
@@ -186,3 +193,11 @@ message HostInitiatorEventDetails {
   // Target uuid
   string uuid = 4;
 }
+
+// Reactor event details
+message ReactorEventDetails {
```
tiagolobocastro, line 207: Frozen reactors can also unfreeze, probably need to take into account here so we can get events for both freeze and unfreeze?

tiagolobocastro: Or maybe just have a ReactorUnfreeze above

datacore-vvarakantham: Added ReactorUnfreeze EventAction now.

tiagolobocastro (approved):

sinhaashish (approved):

PR #59, opened 2024-01-10
fix(apis/events): remove buggy include directive

niladrih:
events build fails on macOS, ref:
https://github.com/openebs/mayastor-extensions/actions/runs/7473781400/job/20338605798


apis/events/src/lib.rs
```diff
@@ -28,7 +28,7 @@ pub mod event {
 // In order to IDE to work properly with event definitions on non-Linux platform,
 // one can copy generated bindings (v1.event.rs) from a Linux target and put them to .pregenerated
 // directory. This has to be done every time event API is changed.
-#[cfg(not(target_os = "linux"))]
-pub mod event {
-    include!("../../../.pregenerated/v1.event.rs");
-}
+// #[cfg(not(target_os = "linux"))]
```
tiagolobocastro, line 31: you also need to remove the cfg for linux only, above on 22

tiagolobocastro (approved):

apis/events/src/lib.rs
niladrih: Done.

abhilashshetty04 (approved):

niladrih:
bors merge

PR #60, opened 2024-01-11
feat(eventing): state change events

datacore-vvarakantham:
Added state change events.

### DIFF
--- a/apis/events/protobuf/v1/event.proto
+++ b/apis/events/protobuf/v1/event.proto
@@ -101,6 +101,8 @@ message EventDetails {
   optional NvmePathEventDetails nvme_path_details = 5;
   // Host initiator event details
   optional HostInitiatorEventDetails host_initiator_details = 6;
+  // State change event details
+  optional StateChangeEventDetails state_change_details = 8;
 }
 
 // Rebuild event details
@@ -185,4 +187,10 @@ message HostInitiatorEventDetails {
   string target = 3;
   // Target uuid
   string uuid = 4;
+}
+
+// State change event details
+message StateChangeEventDetails {
+  string previous = 1;
+  string next = 2;
 }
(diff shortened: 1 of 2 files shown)


apis/events/protobuf/v1/event.proto
```diff
@@ -186,3 +188,9 @@ message HostInitiatorEventDetails {
   // Target uuid
   string uuid = 4;
 }
+
+// State change event details
+message StateChangeEventDetails {
```
tiagolobocastro, line 193: Seems to be quite generic, what is it for?

tiagolobocastro: I see how it's used on the other repos..

tiagolobocastro (approved):

apis/events/src/event_traits.rs
```diff
@@ -177,6 +178,20 @@ impl EventSource {
         }
         self
     }
+
+    /// Add state change event specific data event source.
+    pub fn with_state_change_data(self, previous: &str, next: &str) -> Self {
```
tiagolobocastro: Probably here we should take String, reason being we're most likely converting a type (the state, ex Enum NexusState) to String, so better take String now to avoid 2 conversions!
```suggestion
    pub fn with_state_change_data(self, previous: String, next: String) -> Self {
```
Or
```suggestion
    pub fn with_state_change_data(self, previous: Into<String>, next: Into<String>) -> Self {
```

datacore-vvarakantham: Changed to String.

apis/events/src/event_traits.rs
```diff
@@ -177,6 +178,20 @@ impl EventSource {
...
         self
     }
+
+    /// Add state change event specific data to event source.
+    pub fn with_state_change_data(self, previous: String, next: String) -> Self {
+        EventSource {
+            event_details: Some(EventDetails {
+                state_change_details: Some(StateChangeEventDetails {
+                    previous: previous.to_string(),
+                    next: next.to_string(),
```
tiagolobocastro: previous and next are already String, don't need to to_string again

datacore-vvarakantham: Yes. I removed it now. Thanks

tiagolobocastro (approved):

sinhaashish (approved):

PR #61, opened 2024-01-18
feat(eventing): add event specific details for io-engine stop event

datacore-vvarakantham:
Added event specific details for io-engine stop event.

Added error details to event specific details of nexus shutdown when the nexus shutdown fails.

Added subsystem pause/ resume and nexus init/reconfiguring events.

### DIFF
--- a/apis/events/protobuf/v1/event.proto
+++ b/apis/events/protobuf/v1/event.proto
@@ -4,6 +4,8 @@ import "google/protobuf/wrappers.proto";
 
 import "google/protobuf/timestamp.proto";
 
+import "google/protobuf/duration.proto";
+
 package v1.event;
 
 // Event Message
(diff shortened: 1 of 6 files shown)


tiagolobocastro (approved):

sinhaashish (approved):

apis/events/protobuf/v1/event.proto
```diff
@@ -55,6 +57,10 @@ enum EventAction {
   Shutdown = 18;
   Start = 19;
   Stop = 20;
+  SubsystemPause = 21;
```
tiagolobocastro, line 60: How is the initial pause event sent?

datacore-vvarakantham: I didn't add pause when we receive the pause request. I am creating a pause event when its completed. If there is error I am adding error to the event details. Else I am just adding duration for the pause action. So if there is no error it is assumed to be successful.

datacore-vvarakantham: Or should I add an event when pause request is received? And another event when the pause fails?

tiagolobocastro: Yes, 1 when it's started, and 1 when it's finished.
Reason for this is that pause sometimes gets stuck, in which case we'd never get the event, so would be good to know if a pause was started at some point or not.

datacore-vvarakantham: Added the events when subsystem pause started and completed.

tiagolobocastro:
Can his be merged @datacore-vvarakantham ?

datacore-vvarakantham:
> Can his be merged @datacore-vvarakantham ?

Yes, it can be merged.

PR #62, opened 2024-01-19
feat(eventing): snapshot and clone events

datacore-vvarakantham:
Added snapshot and clone events.

### DIFF
--- a/apis/events/protobuf/v1/event.proto
+++ b/apis/events/protobuf/v1/event.proto
@@ -32,6 +32,8 @@ enum EventCategory {
   NvmePath = 7;
   HostInitiator = 8;
   IoEngineCategory = 9;
+  Snapshot = 10;
+  Clone = 11;
 }
 
 // Event Action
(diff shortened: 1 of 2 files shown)


apis/events/protobuf/v1/event.proto
```diff
@@ -210,3 +216,21 @@ message ReactorEventDetails {
   // Reactor state
   string state = 2;
 }
+
+// Snapshot event details
+message SnapshotEventDetails {
+  // Parent id of the snapshot
+  string parent_id = 1;
```
tiagolobocastro: what is the parent id?

datacore-vvarakantham: Parent_id is replica_id. And entity_id is volume_id. Should I rename the variables?

tiagolobocastro: Yes I think would be helpful here

datacore-vvarakantham: Changed the variable names to replica_id and volume_id.

tiagolobocastro (approved):

sinhaashish (approved):

PR #64, opened 2024-02-01
fix(build): fixing issue with pregenerated code on VSCode

dsavitskiy:
VSCode IDE does not seem to parse properly include! macro for upper dirs


apis/io-engine/src/v0.rs
```diff
@@ -36,7 +36,7 @@ include!(concat!(env!("OUT_DIR"), "/mayastor.rs"));
 // This has to be done every time Mayastor API V0 is changed.
 #[cfg(not(target_os = "linux"))]
 mod v0_generated {
-    include!("../../../.pregenerated/mayastor.rs");
+    include!(".pregenerated/mayastor.rs");
```
tiagolobocastro: Is this really necessary? Doesn't mac generate this too?
I remember something like this was causing an issue before on the macos builds on github actions.

tiagolobocastro (approved):

dsharma-dc (approved):

PR #65, opened 2024-02-12
fix(devinfo): add consistent read for kernel bug 

### LINKED ISSUE #1274: csi-node is unable to find mount during statefulset scale in
**Describe the bug**
We are running some scale tests with mayastor and mayastor-control-plane on AKS. When the number of pods per node is large(~80), there are some issues while scaling in and the pods remain in Terminating state.

This happens because, the find_mount check [here ](https://github.com/openebs/mayastor-control-plane/blob/develop/control-plane/csi-driver/src/bin/node/filesystem_vol.rs#L322)returns false, even though the target mount is there in the /proc/mounts file.
As a result, we run into : "Device or resource busy (os error 16)" errors on the csi-node when the csi-node tries to remove the directory here: 
https://github.com/openebs/mayastor-control-plane/blob/develop/control-plane/csi-driver/src/bin/node/filesystem_vol.rs#L326

**To Reproduce**

**Expected behavior**
find_mount should return true when the mount record is present in /proc/mounts

**Screenshots**

** OS info (please complete the following information):**
 - Distro: [e.g. NixOS]
 - Kernel version
 - MayaStor revision or container image

**Additional context**
Related PR: https://github.com/openebs/mayastor-control-plane/pull/413

niladrih:
Fixes https://github.com/openebs/mayastor/issues/1274

On the linux kernel versions below v5.8, there's a bug which behaves like so:
Whenever a read(2) syscall is issued (and this one takes an argument for how many bytes to read off of a file) to the `/proc/mounts` virtual file, few of the mount entries are read off of it. The kernel maintains the seek position of the file so that it can serve subsequent read(2)s from that position. But, in between the first and the second read(2), a device might be unmounted and one/few mount entries might be removed from the already-read portion of the file. This moves the mount entries after the deleted entries further up the file. However, the seek position has not changed. So, subsequent read(2)s will miss one or more read entries, as the seek position now points to a mount entry further down the list.
Ref: https://github.com/torvalds/linux/commit/9f6c61f96f2d97cbb5f7fa85607bc398f843ff0f

Changes:
This PR adds a `consistent_read()` implementation for the interface `MountIter` for kernel versions less than 5.8.0. `consistent_read()` reads the whole file twice start to end, and compares the two byte buffers. If both of the subsequent reads resulted in the same byte-sequence, then we can say we have not lost mount entries when traversing `/proc/mounts`, as then the comparison would fail on the second read from start to end. This is configured to default to 2 comparison retries if the first comparison fails. However that may not be enough if there are very many mount entries and the host has very frequent unmounts. Use higher retries for those cases.

### DIFF
--- a/devinfo/src/mountinfo/io_utils.rs
+++ b/devinfo/src/mountinfo/io_utils.rs
@@ -1,0 +1,30 @@
+use crate::mountinfo::error::{MountInfoError, Result};
+use std::{fs::read, path::Path};
+
+const DEFAULT_RETRY_COUNT: u32 = 2;
+
+/// Issue std::fs::read on a file twice and compare the two byte sequences
+/// to see if the reads were consistent. Retry if the reads resulted in different byte
+/// sequences.
+pub(crate) fn consistent_read<P: AsRef<Path>>(
+    path: P,
+    retry_count: Option<u32>,
+) -> Result<Vec<u8>> {
+    let mut current_content = read(path.as_ref())?;
+
+    let retries = retry_count.unwrap_or(DEFAULT_RETRY_COUNT);
+    for _ in 0 ..= retries {
+        let new_content = read(path.as_ref())?;
+
+        if new_content.eq(&current_content) {
+            return Ok(new_content);
+        }
+
+        current_content = new_content;
+    }
+
+    Err(MountInfoError::InconsistentRead {
+        filepath: path.as_ref().to_path_buf(),
+        retries,
+    })
+}
--- a/devinfo/Cargo.toml
+++ b/devinfo/Cargo.toml
@@ -5,6 +5,8 @@ authors = ["Jeffry Molanus <jeffry.molanus@gmail.com>"]
 edition = "2018"
 
 [dependencies]
+nix = { version = "0.27.1", default-features = false, features = [ "feature" ] }
+semver = "1.0.20"
 snafu = "0.7.5"
 url = "2.4.1"
 uuid = { version = "1.4.1", features = ["v4"] }
--- a/devinfo/src/mountinfo/bytebuf.rs
+++ b/devinfo/src/mountinfo/bytebuf.rs
@@ -1,0 +1,41 @@
+use std::io::{IoSliceMut, Read};
+
+/// This is io::Read capable Sized type. This can wrap around a Vec<u8>.
+pub struct ByteBuf {
+    inner: Vec<u8>,
+}
+
+impl Read for ByteBuf {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        self.inner.as_slice().read(buf)
+    }
+
+    fn read_vectored(&mut self, bufs: &mut [IoSliceMut<'_>]) -> std::io::Result<usize> {
+        self.inner.as_slice().read_vectored(bufs)
+    }
+
+    fn read_to_end(&mut self, buf: &mut Vec<u8>) -> std::io::Result<usize> {
+        self.inner.as_slice().read_to_end(buf)
+    }
+
+    fn read_to_string(&mut self, buf: &mut String) -> std::io::Result<usize> {
+        self.inner.as_slice().read_to_string(buf)
+    }
+
+    fn read_exact(&mut self, buf: &mut [u8]) -> std::io::Result<()> {
+        self.inner.as_slice().read_exact(buf)
+    }
+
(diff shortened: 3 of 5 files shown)


niladrih:
@tiagolobocastro and others -- should I remove the MountIter::new() API as I intend to refactor control-plane and mayastor to use SafeMountIter instead?

niladrih:
Is the linter CI issue because of std::sync::OnceLock?

tiagolobocastro:
> Is the linter CI issue because of std::sync::OnceLock?

Non-related, @datacore-vvarakantham is fixing this

tiagolobocastro:
> @tiagolobocastro and others -- should I remove the MountIter::new() API as I intend to refactor control-plane and mayastor to use SafeMountIter instead?

We need to keep I think, otherwise you can't merge this PR independently to the other repos?

niladrih:
> > @tiagolobocastro and others -- should I remove the MountIter::new() API as I intend to refactor control-plane and mayastor to use SafeMountIter instead?
> 
> We need to keep I think, otherwise you can't merge this PR independently to the other repos?

Yeah, it's fine to keep it.

devinfo/src/mountinfo/io_utils.rs
```diff
@@ -0,0 +1,27 @@
...
+const DEFAULT_RETRY_COUNT: u32 = 2;
+
+pub(crate) fn consistent_read<P: AsRef<Path>>(
+    path: P,
+    retry_count: Option<u32>,
+) -> Result<Vec<u8>> {
+    let mut current_content = read(path.as_ref())?;
+
+    let retries = retry_count.unwrap_or(DEFAULT_RETRY_COUNT);
+    for _ in 0 .. retries {
```
tiagolobocastro: Otherwise there's no check? Or was this intended?

```suggestion
    for _ in 0 ..= retries {
```

devinfo/src/mountinfo/io_utils.rs
```diff
@@ -0,0 +1,27 @@
+use crate::mountinfo::error::{MountInfoError, Result};
+use std::{fs::read, path::Path};
+
+const DEFAULT_RETRY_COUNT: u32 = 2;
+
+pub(crate) fn consistent_read<P: AsRef<Path>>(
+    path: P,
+    retry_count: Option<u32>,
```
tiagolobocastro, line 11: You are specifying Some(2), does this need to be optional?

devinfo/src/mountinfo/io_utils.rs
niladrih: This is intentional. Taking an example: retries=2,
0..= retries goes from 0 to 2, which is 3 iterations
0.. retries goes to 0 to 1, which is 2 iterations, which is what we want when we say 2 retries.

Does this make sense? Or is retries in the sense that the usual try and then the retries, then it comes to 3.

devinfo/src/mountinfo/io_utils.rs
niladrih: Sure, why not..

devinfo/src/mountinfo/io_utils.rs
niladrih: I've renamed the variable to 'attempts'.

niladrih: > Otherwise there's no check 

Oh I get your point. Yeah, needs at least 1. Needs '='.

niladrih: Renamed back to retries, makes more sense now with the '='.

devinfo/src/mountinfo/io_utils.rs
niladrih: Added the `SafeMountIter::get_with_retries(retry_count)` API, which takes an optional argument for the retry-count.
Changed `SafeMountIter::get()` to use `SafeMountIter::get_with_retries()` so that it reverts to the default retry count of 2 when that API is used.

tiagolobocastro (approved):

devinfo/src/mountinfo/io_utils.rs
tiagolobocastro: > > Otherwise there's no check
> 
> Oh I get your point. Yeah, needs at least 1. Needs '='.

Yeah, for retries = 2:
0..=2
0: compare first time
1: retry n1
2: retry n2

Abhinandan-Purkait (approved):

PR #68, opened 2024-02-18
refactor(devinfo): make mountinfo APIs more accessable to caller

niladrih:
Changes:
- Makes devinfo::mountinfo::error module 'pub' so that callers may access errors from this module.
- Makes the SAFE_MOUNT_ITER `OnceLock<SafeMountIter>`, ByteBuf struct and `MountIter<R>::new_from_readable()` private, as they aren't required outside of this module.

### DIFF
--- a/devinfo/src/mountinfo/bytebuf.rs
+++ b/devinfo/src/mountinfo/bytebuf.rs
@@ -1,7 +1,7 @@
 use std::io::{IoSliceMut, Read};
 
-/// This is io::Read capable Sized type. This can wrap around a Vec<u8>.
-pub struct ByteBuf {
+/// This is io::Read capable Sized type. This wraps around a Vec<u8>.
+pub(crate) struct ByteBuf {
     inner: Vec<u8>,
 }
 
--- a/devinfo/src/mountinfo/mod.rs
+++ b/devinfo/src/mountinfo/mod.rs
@@ -17,8 +17,11 @@ use std::{
     sync::OnceLock,
 };
 
+/// Contains a type which is Sized and implements io::Read.
+/// Useful as a Read-able alternative to a Vec<u8>.
 mod bytebuf;
-mod error;
+/// Errors for MountInfo affairs.
+pub mod error;
 /// Contains tools to interact with files, etc.
 mod io_utils;
 
@@ -226,7 +229,7 @@ impl<R: BufRead> Iterator for MountIter<R> {
     }
 }
 
-pub static SAFE_MOUNT_ITER: OnceLock<SafeMountIter> = OnceLock::new();
+static SAFE_MOUNT_ITER: OnceLock<SafeMountIter> = OnceLock::new();
 
 /// This returns a Result<Iterator> with reads /proc/mounts consistently.
 pub struct SafeMountIter {


devinfo/src/mountinfo/io_utils.rs
```diff
@@ -6,10 +6,7 @@ const DEFAULT_RETRY_COUNT: u32 = 2;
 /// Issue std::fs::read on a file twice and compare the two byte sequences
 /// to see if the reads were consistent. Retry if the reads resulted in different byte
 /// sequences.
-pub(crate) fn consistent_read<P: AsRef<Path>>(
-    path: P,
-    retry_count: Option<u32>,
-) -> Result<Vec<u8>> {
+pub fn consistent_read<P: AsRef<Path>>(path: P, retry_count: Option<u32>) -> Result<Vec<u8>> {
```
tiagolobocastro: what is the reason to make this accessible directly?

niladrih: No reason. My primary issue was that the 'error' module was private. No real reason for this to be public, and I see no other use-case.

niladrih: Made it private again. Made the module io_utils private as well.

Abhinandan-Purkait (approved):

devinfo/src/mountinfo/mod.rs
```diff
@@ -17,8 +17,11 @@ use std::{
     sync::OnceLock,
 };
 
-mod bytebuf;
-mod error;
+/// Contains a type which is Sized and implements io::Read.
+/// Useful as a Read-able alternative to a Vec<u8>.
+pub mod bytebuf;
```
Abhinandan-Purkait: Where are we using this?

niladrih: In this moddule, i.e. mountinfo, we are using this in the `get_with_retries()` method. Outside this module, it's not used anywhere. But it seemed like useful tool which one might want to use outside of this context. Wdyt?

tiagolobocastro: I'd say keep all private unless there's a specific reason to open up

devinfo/src/mountinfo/bytebuf.rs
```diff
@@ -1,6 +1,7 @@
 use std::io::{IoSliceMut, Read};
 
-/// This is io::Read capable Sized type. This can wrap around a Vec<u8>.
+/// This is io::Read capable Sized type. This wraps around a Vec<u8>.
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct ByteBuf {
```
tiagolobocastro: Does this need to be pub?

niladrih: Made this private. Also removed the `#[derive(Clone, Debug, Eq, PartialEq)]` as these are not required for this crate's use.

devinfo/src/mountinfo/mod.rs
niladrih: Made this module private.

tiagolobocastro (approved):

PR #71, opened 2024-02-19
Add Lvm backend support

tiagolobocastro:


apis/io-engine/protobuf/v1/replica.proto
```diff
@@ -45,6 +47,7 @@ message Replica {
   bool is_snapshot = 11; // true: snapshot false: normal replica
   bool is_clone = 12; // true: clone replica from snapshot, false: normal replica
   optional string snapshot_uuid = 13; // snapshot uuid from which the clone replica is created
+  mayastor.v1.PoolType pooltype = 14; // type of the pool (eg: lvm, lvm)
```
hrudaya21: Typo in the comment, lvm mentioned twice. Need to fix in all places in this file.

hrudaya21 (approved):

apis/io-engine/protobuf/v1/replica.proto
tiagolobocastro: It's to emphasize lvm :D
I've also been wondering if instead of specifying type like this we should add oneof for the create requests, example:
```
oneof pooltype {
    LvmArgs lvm = 1;
    LvsArgs lvs = 2;
}
```
wdyt?

hrudaya21: Yes, it more sense. As We can have just one type for one pool.

tiagolobocastro: Sadly seems existing pool api already uses the type and not oneof, so probably better to do the same for consistency :(

apis/io-engine/protobuf/v1/replica.proto
```diff
@@ -55,6 +58,7 @@ message CreateReplicaRequest {
   bool thin = 5;    // thin provisioning
   ShareProtocol share = 6;  // protocol to expose the replica over
   repeated string allowed_hosts = 7; // host (nqn's) which are allowed to connect to the target.
+  mayastor.v1.PoolType pooltype = 8; // type of the pool (eg: lvm, lvm)
```
Abhinandan-Purkait: lvm, lvs ?

Abhinandan-Purkait: At other places as well?

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

dsavitskiy (approved):

PR #72, opened 2024-02-21
feat(devinfo): implement From<MountInfoError> for std::io::Error

niladrih:


devinfo/src/mountinfo/error.rs
```diff
@@ -65,3 +65,9 @@ impl From<semver::Error> for MountInfoError {
         Self::Semver(semver_err)
     }
 }
+
+impl From<MountInfoError> for std::io::Error {
+    fn from(err: MountInfoError) -> Self {
+        std::io::Error::new(std::io::ErrorKind::Other, err)
```
tiagolobocastro: should we match on the various error types and build a more precise error?

niladrih: Oh, I'm sorry that was lazy of me. Sure!

devinfo/src/mountinfo/error.rs
```diff
@@ -65,3 +65,17 @@ impl From<semver::Error> for MountInfoError {
...
     }
 }
+
+impl From<MountInfoError> for std::io::Error {
+    fn from(error: MountInfoError) -> Self {
+        use std::io::{Error, ErrorKind};
+
+        match error {
+            MountInfoError::Io(std_io_err) => std_io_err,
+            MountInfoError::InconsistentRead { .. } => Error::new(ErrorKind::Other, error),
```
tiagolobocastro: Maybe busy better to explain the inconsistent read? 

```suggestion
            MountInfoError::InconsistentRead { .. } => Error::new(ErrorKind::ResourceBusy, error),
```

tiagolobocastro (approved):

devinfo/src/mountinfo/error.rs
tiagolobocastro: Yeah this was the whole reason why I suggest to do this, so we can build a "better" error

devinfo/src/mountinfo/error.rs
niladrih: Hmm, yeah that actually fits better.

niladrih: ResourceBusy is an unstable feature at the moment unfortunately.

tiagolobocastro: Damn, maybe InvalidData then

tiagolobocastro: Or whichever one you prefer then

niladrih: Sure, using InvalidData. Putting in a TODO for changing it some time in the future.

abhilashshetty04 (approved):

PR #75, opened 2024-03-13
feat(callhome): add bustype and rotational properties to block device

datacore-vvarakantham:
Added bus type("connection_type") and rotational ("is_rotational") properties to block device to be able to generate storage media metrics.

connection_type: the type of bus through which the device is connected to the system
is_rotational: indicates whether the device is rotational or non-rotational

### DIFF
--- a/apis/io-engine/protobuf/v1/host.proto
+++ b/apis/io-engine/protobuf/v1/host.proto
@@ -56,6 +56,8 @@ message BlockDevice {
   Partition partition = 9;      // partition information in case where device represents a partition
   Filesystem filesystem = 10;   // filesystem information in case where a filesystem is present
   bool available = 11;          // identifies if device is available for use (ie. is not "currently" in use)
+  string connection_type = 12;   // the type of bus through which the device is connected to the system
+  optional bool is_rotational = 13;     // indicates whether the device is rotational or non-rotational
 }
 
 message ListBlockDevicesRequest {


apis/io-engine/protobuf/mayastor.proto
tiagolobocastro, line 1: There's no need to modify the v0 api

apis/io-engine/protobuf/v1/host.proto
```diff
@@ -56,6 +56,8 @@ message BlockDevice {
   Partition partition = 9;      // partition information in case where device represents a partition
   Filesystem filesystem = 10;   // filesystem information in case where a filesystem is present
   bool available = 11;          // identifies if device is available for use (ie. is not "currently" in use)
+  string connectiontype = 12;   // the type of bus through which the device is connected to the system
```
tiagolobocastro:
```suggestion
  string connection_type = 12;   // the type of bus through which the device is connected to the system
```

apis/io-engine/protobuf/v1/host.proto
```diff
@@ -56,6 +56,8 @@ message BlockDevice {
...
+  string isrotational = 13;     // indicates whether the device is rotational or non-rotational (1 or 0)
```
tiagolobocastro:
```suggestion
  bool is_rotational = 13;     // indicates whether the device is rotational or non-rotational
```

apis/io-engine/protobuf/mayastor.proto
datacore-vvarakantham: Removed it from the v0 api.

apis/io-engine/protobuf/v1/host.proto
tiagolobocastro: resolved but it's still string...

datacore-vvarakantham: Yes. The queue/rotational attribute may or may not be present. Should I make it optional bool ?

tiagolobocastro: Yes that makes sense

datacore-vvarakantham: Made it optional bool

tiagolobocastro (approved):

sinhaashish (approved):

PR #80, opened 2024-04-26
feat(nvmeadm): allow overriding the hostid and the hostnqn prefix

tiagolobocastro:
Also set default hostnqn to match the one set by nvme-cli. Recent kernels are more strict and don't allow mixing the initial connect with hostid/hostnqn with another.

### DIFF
--- a/nvmeadm/src/nvmf_discovery.rs
+++ b/nvmeadm/src/nvmf_discovery.rs
@@ -28,7 +28,12 @@ use crate::{
 /// referred as.
 const MACHINE_UUID_PATH: &str = "/sys/class/dmi/id/product_uuid";
 
-static HOST_ID: once_cell::sync::Lazy<String> = once_cell::sync::Lazy::new(|| {
+struct HostInfo {
+    id: String,
+    nqn: String,
+    nqn_prefix: &'static str,
+}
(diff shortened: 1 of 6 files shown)


Abhinandan-Purkait (approved):

nvmeadm/src/nvmf_discovery.rs
```diff
@@ -39,7 +44,12 @@ static HOST_ID: once_cell::sync::Lazy<String> = once_cell::sync::Lazy::new(|| {
             }
         }
     }
-    host_id
+    let nqn_prefix = "nqn.2014-08.org.nvmexpress:uuid";
```
dsharma-dc, line 47: If I see this in conjunction with our nqn discussion today, why do we choose to use prefix that's usually on initiator side(linux kernel)? Currently it is `nqn.2019-05.io.openebs.engine` few lines below. I don't know this code completely so may be missing some point.

tiagolobocastro: Because newer kernels no longer allow multiple combinations of nqn to id
This allows us to use nvme connect whilst testing, without having to specify hostid.
For "production control-plane" we specify the hostnqn rather than take default.

dsharma-dc (approved):

PR #82, opened 2024-05-10
ci: add config.nix file for storing shared constants

niladrih:


niladrih:
This change is not required at the moment.

PR #85, opened 2024-05-24
refactor(snapshot/rebuild/api): allow for push mode

tiagolobocastro:
Allows for separate rebuild modes: pull/push
This means we can also run the rebuild job from the snapshot node in push.

### DIFF
--- a/apis/io-engine/protobuf/v1/snapshot-rebuild.proto
+++ b/apis/io-engine/protobuf/v1/snapshot-rebuild.proto
@@ -6,19 +6,35 @@ import "google/protobuf/duration.proto";
 
 package mayastor.v1;
 
+// Snapshot Rebuild job which copies from a snapshot into a replica.
+// If this service is supported then the `snapshotRebuild` feature
+// must be enabled in the HostRpc::MayastorFeatures.
 service SnapshotRebuildRpc {
+  // Create and start the rebuild job itself.
   rpc CreateSnapshotRebuild(CreateSnapshotRebuildRequest) returns (SnapshotRebuild) {}
+  // List all snapshot rebuild jobs.
   rpc ListSnapshotRebuild(ListSnapshotRebuildRequest) returns (ListSnapshotRebuildResponse) {}
+  // Destroys a snapshot rebuild job.
   rpc DestroySnapshotRebuild(DestroySnapshotRebuildRequest) returns (google.protobuf.Empty) {}
 }
 
 message CreateSnapshotRebuildRequest {
+  // The uuid of the rebuild.
+  string                             uuid = 1;
   // The uuid of the rebuilding replica.
-  string                     replica_uuid = 1;
-  // The nvmf URI of the source Snapshot from which we:
+  string                     replica_uuid = 2;
+  // The uuid of the source snapshot.
+  string                    snapshot_uuid = 3;
+
+  // The URI of the destination replica.
+  // If empty, then it's a local replica.
+  string                      replica_uri = 4;
+  // The URI of the source Snapshot from which we:
   // 1. read allocated data bitmap
   // 2. read data proper
-  string                       source_uri = 2;
+  // If empty, then it's a local snapshot.
+  string                     snapshot_uri = 5;
+
   // In case of crash/restart, we can first
   // 1. read allocated data bitmap
   // and then depending on the resume flag:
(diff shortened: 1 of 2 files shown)


apis/io-engine/protobuf/v1/snapshot-rebuild.proto
```diff
@@ -29,10 +41,12 @@ message CreateSnapshotRebuildRequest {
   bool                             resume = 3;
   // A bitmap where set bits indicate which
   // segment of the snapshot are allocated and therefore
-  // need to be transferred for the rebuild.
+  // need to be transferred from the source to rebuild.
   // If not specified the replica shall retrieve it
   // using nvme-admin command.
   optional SnapshotRebuildBitmap   bitmap = 4;
+  /// When running on the snapshot node, use bitmap mode as well.
+  bool                       local_bitmap = 8;
```
dsharma-dc: Do we really have to provide a bool switch to use or not-use bitmap? Shall we not simply use bitmap internally if it is present?

dsharma-dc (approved):

apis/io-engine/protobuf/v1/snapshot-rebuild.proto
tiagolobocastro: This one is a bit awkward, perhaps naming is not entirely correct.
Basically we can either receive the bitmap via gRPC and walk it, or we can retrieve the bitmap and walk, or we can just walk the device.
Maybe this should be an oneof instead,
oneof {
SnapshotRebuildBitmap.
FetchBitmap
NoBitmap,
} 
Suggestions?

dsharma-dc: > receive the bitmap via gRPC and walk it

Receive via gRPC from where? - Considering a case say source snapshot is of a replica on node1, and replica rebuild destination is on node2.

tiagolobocastro: Bitmap can be specified via gRPC parameters, L47

dsharma-dc: ah right. `oneof` would be nice. Something like...

```
{
bitmap_consumer_remote,
bitmap_consumer_local,
no_bitmap
}
```

tiagolobocastro: hmm doesn't didn't quite work here as it expects each one to be a field, so we'd have to add fields, ex:
```
  oneof RebuildBitmapArgs {
    int NoBitmap = 1;
    SnapshotRebuildBitmap SpecifiedArgs = 2;
    int NvmeAdmin = 3;
  }
```

Just the type SnapshotRebuildBitmap with a bool is probably simpler:
```
  // A bitmap where set bits indicate which
  // segment of the snapshot are allocated and therefore
  // need to be transferred from the source to rebuild.
  // If not specified the replica shall retrieve it
  // using nvme-admin command.
  optional SnapshotRebuildBitmap   bitmap = 4;
  // Use the bitmap for the rebuild, or just walk the device range.
  // If the bitmap is specified via the args use it, otherwise fetch
  // it from the snapshot.
  bool                         use_bitmap = 8;
```

niladrih (approved):
Makes sense to me :P

apis/io-engine/protobuf/v1/snapshot-rebuild.proto
```diff
@@ -29,10 +43,14 @@ message CreateSnapshotRebuildRequest {
...
   // segment of the snapshot are allocated and therefore
-  // need to be transferred for the rebuild.
+  // need to be transferred from the source to rebuild.
   // If not specified the replica shall retrieve it
   // using nvme-admin command.
   optional SnapshotRebuildBitmap   bitmap = 4;
+  // Use the bitmap for the rebuild, or just walk the device range.
+  // If the bitmap is specified via the args use it, otherwise fetch
+  // it from the snapshot.
+  bool                         use_bitmap = 8;
```
dsharma-dc, line 55: Why is this field numbered `8` ? There is no `7`.Also, I'm not sure of reason of numbering fields non-sequentially - does it help here in structure packing efficiently and reduce padding?

apis/io-engine/protobuf/v1/snapshot-rebuild.proto
dsharma-dc: ok, let's keep it as is then.

Abhinandan-Purkait (approved):

apis/io-engine/protobuf/v1/snapshot-rebuild.proto
tiagolobocastro: It's for backwards compatibility, though since we never released I guess it's fine to start from 1-onwards

PR #86, opened 2024-06-07
feat: expose nexus rebuild fix

tiagolobocastro:
This is useful to ensure nexus rebuild is done correctly, copying all the clusters from an existing replica.

### DIFF
--- a/apis/io-engine/protobuf/v1/common.proto
+++ b/apis/io-engine/protobuf/v1/common.proto
@@ -7,3 +7,17 @@ enum ShareProtocol {
   NVMF = 1; // NVMe over Fabrics (TCP)
   ISCSI = 2; // iSCSI
 }
+
+message MayastorFeatures {
+  // NVMe ANA enablement status.
+  bool asymmetricNamespaceAccess = 1;
+  // LVM backend presence and enablement status.
+  optional bool logicalVolumeManager = 2;
+  // SnapshotRebuild presence and enablement status.
+  optional bool snapshotRebuild = 3;
+}
+
+message MayastorBugFixes {
+  // Nexus rebuilds both the clusters allocated to the replica and its ancestors clusters
+  bool nexusRebuildReplicaAncestry = 1;
+}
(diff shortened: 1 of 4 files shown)


apis/io-engine/protobuf/v1/common.proto
```diff
@@ -7,3 +7,17 @@ enum ShareProtocol {
   NVMF = 1; // NVMe over Fabrics (TCP)
   ISCSI = 2; // iSCSI
 }
+
+message MayastorFeatures {
+  // NVMe ANA enablement status.
+  bool asymmetricNamespaceAccess = 1;
+  // LVM backend presence and enablement status.
```
abhilashshetty04, line 14: @tiagolobocastro , shouldnt this be snake case?

apis/io-engine/protobuf/v1/host.proto
```diff
@@ -20,16 +20,15 @@ service HostRpc {
   rpc StatNvmeController (StatNvmeControllerRequest) returns (StatNvmeControllerResponse) {}
 }
 
-message MayastorFeatures {
+message BackCompatMayastorFeatures {
+  // NVMe ANA enablement status.
   bool asymmetricNamespaceAccess = 1;
```
dsharma-dc, line 25: Why do we need separate `BackCompatMayastorFeatures` ?

tiagolobocastro: Just for back-compat for existing code checking the old features directly from HostInfo, which is just 1 feature: ANA

apis/io-engine/protobuf/v1/common.proto
tiagolobocastro: Existing type before I moved it was like that if you see, so I kept in.
Doesn't make a difference to generated code though, this is just on the proto.
I can change to snake though if you like?

apis/io-engine/protobuf/v1/host.proto
tiagolobocastro: I guess another option could be to add a get_info_v2 and use new types for the whole thing?

dsharma-dc (approved):

Abhinandan-Purkait (approved):

PR #88, opened 2024-07-30
build: add common release script

tiagolobocastro:
The script header showcases how to use it.
This will allow other repos to simply make use of this script and add unique requirements on top of it.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,0 +1,623 @@
+#!/usr/bin/env bash
+
+# A helper script to Build and upload mayastor docker images to dockerhub repository.
+# Use --dry-run to just see what would happen.
+# The script assumes that a user is logged on to dockerhub for public images,
+# or has insecure registry access setup for CI.
+
+# Another script must source this script, with the given env variables:
+# PROJECT: name of the project (useful for dep caching)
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -0,0 +1,417 @@
...
+        image="${REGISTRY}/${image}"
+      fi
+    fi
+    # If we're skipping the build, then we just want to upload
+    # the images we already have locally.
+    if [ -z $SKIP_BUILD ]; then
+      archive=${name}
+      echo "Building $image:$TAG ..."
+      $NIX_BUILD --out-link $archive-image -A images.$BUILD_TYPE.$archive --arg allInOne "$ALL_IN_ONE" --arg incremental "$INCREMENTAL" --argstr product_prefix "$PRODUCT_PREFIX"
+      $DOCKER load -i $archive-image
```
niladrih: Should it be possible to disable load with a switch?

tiagolobocastro: Yes, that's part of the reason that made me want to do this. Will add skopeo support and make loading optional.

niladrih: Okay.

niladrih (approved):
The extensions script, today, has a helm chart script section. Nothing to do here, but an FYI.

niladrih (approved):

Abhinandan-Purkait (approved):

PR #91, opened 2024-08-19
feat(pool): adding support for pool grow

dsavitskiy:


apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -33,14 +34,34 @@ message PoolTypeValue {
...
+  // This hint is ratio of the desired to the pool backend's default reservation.
+  //
+  // 1 is the default value, and it means to use the pool backend's default
+  // reservation amount.
+  //
+  // 0.5 means an attempt to reserve two times less capacity for the metadata than
+  // the backend would do by default, trading replica/snapshot count
+  // for the user data space.
+  //
+  // 2 means a battempt to reserve two times more space for the metadata than
```
```suggestion
  // 2 means an attempt to reserve two times more space for the metadata than
```

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -61,15 +82,46 @@ enum PoolState {
...
+  // Number of metadata pages used for storing metadata for the currently
+  // existing replicas.
+  // If the pool backend does not support metadata capacity evaluation, zero is returned.
+  uint64 md_used_pages = 13;
+  // Usable capacity of the pool's underlying storage device(s).
+  // It changes when the device's capacity changes.
+  // This value is equal or large than the current pool capacity. If it is larger,
+  // pool can be grown to fill the entire device's capacity.
+  // If the pool backend does not support disk capacity evaluation, zero is returned.
+  uint64 disk_capacity = 14;
```
tiagolobocastro: We could move this to another Message?
If the Message is Some is sent then it's new version, otherwise it's old version?

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -33,14 +34,34 @@ message PoolTypeValue {
...
+  // 0.5 means an attempt to reserve two times less capacity for the metadata than
+  // the backend would do by default, trading replica/snapshot count
+  // for the user data space.
+  //
+  // 2 means a battempt to reserve two times more space for the metadata than
+  // the backend would do by default, allowing pool to accomodate more
+  // replicas/snapshots by reducing user data capacity.
+  //
+  // A pool backend is allowed to ignore this hint.
+  optional float md_resv_hint = 6;
```
tiagolobocastro: Is it possible that we set this value too high?
If so will return error code let us know this?

dsavitskiy: My idea is this is "hint" kind of parameter: a backend is free to clamp to a usable value, or ignore it.
We can change this spec to that a backend _may_, or _must_ fail with EINVAL if the value is not usable. I'm not sure which is better here, _may_ or _must_.

tiagolobocastro: Ah you mean, if it's too large it just clamps it to the max it can for example? Seems acceptable I think.
What do you think as well @dsharma-dc

apis/io-engine/protobuf/v1/pool.proto
dsavitskiy: Done

apis/io-engine/protobuf/v1/pool.proto
dsavitskiy: I moved this MD arg into a separate optional message, and renamed it md_resv_ratio.

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -33,14 +34,41 @@ message PoolTypeValue {
...
+  // Values larger that 1.0 would increase amount of space reserved for metadata.
+  // For example, 2.0 would double the capacity reserved for metadata.
+  //
+  // Values smaller that 1.0 would decrease amount of space reserved for metadata.
+  // For example, 0.5 would half the capacity reserved for metadata.
+  //
+  // The actual amount of disk space in bytes reserved for metadata may
+  // depend on pool size, cluster size, and other pool parameters.
+  // A pool backend may also allow dynamic metadata allocation.
+   optional float md_resv_ratio = 1;
```
tiagolobocastro, line 71: Does this still need to be optional if the message is already optional?

tiagolobocastro (approved):

apis/io-engine/protobuf/v1/pool.proto
dsharma-dc: Clamping sounds reasonable, but the question would be how high is `too high`. Instead of providing this hint in terms of metadata, I would suggest let's provide it in terms of `how much we expect pool to grow`? For example, If API field is set to 2, meaning we expect a 2x growth of lvstore in future, so we directly map this 2 as 200 into `num_md_pages_per_cluster_ratio` for spdk.

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -61,15 +89,52 @@ enum PoolState {
...
+  // Page size in bytes. This is the write and read granularity.
+  uint32 page_size = 10;
+  // Usable capacity of the pool's underlying storage device(s).
+  // It changes when the device's capacity changes.
+  // This value is equal or large than the current pool capacity. If it is larger,
+  // pool can be grown to fill the entire device's capacity.
+  // If the pool backend does not support disk capacity evaluation, zero is returned.
+  uint64 disk_capacity = 11;
+  // Optional pool medata info.
+  optional PoolMetadataInfo md_info = 12;
```
dsharma-dc, line 120: Why do we need to expose this info in the API? Is this of any use to API consumer?

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -61,15 +89,52 @@ enum PoolState {
...
+  uint32 cluster_size = 9;
+  // Page size in bytes. This is the write and read granularity.
+  uint32 page_size = 10;
+  // Usable capacity of the pool's underlying storage device(s).
+  // It changes when the device's capacity changes.
+  // This value is equal or large than the current pool capacity. If it is larger,
+  // pool can be grown to fill the entire device's capacity.
+  // If the pool backend does not support disk capacity evaluation, zero is returned.
+  uint64 disk_capacity = 11;
+  // Optional pool medata info.
```
dsharma-dc:
```suggestion
  // Optional pool metadata info.
```

apis/io-engine/protobuf/v1/pool.proto
dsavitskiy: I think yes: to tell the backend to use the default value.

apis/io-engine/protobuf/v1/pool.proto
dsavitskiy: Tiago wanted this for the control plane, so it can decide if a pool should be expanded or not. If the pool approches its metadata capacity limit, control plane may decide not to grow.

apis/io-engine/protobuf/v1/pool.proto
dsavitskiy: Fixed

dsharma-dc (approved):

dsavitskiy:
bors merge

PR #92, opened 2024-08-29
chore: implement FromStr trait for TrType

dsharma-dc:
This will be used to decide TrType variant based on Attach trait values in the nvmf connect/attachment path.

### DIFF
--- a/nvmeadm/src/nvmf_discovery.rs
+++ b/nvmeadm/src/nvmf_discovery.rs
@@ -73,6 +73,19 @@ impl fmt::Display for TrType {
     }
 }
 
+impl FromStr for TrType {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "rdma" => Ok(Self::rdma),
+            "tcp" => Ok(Self::tcp),
+            "fc" => Ok(Self::fc),
+            invalid => Err(format!("Invalid transport type: {invalid}")),
+        }
+    }
+}
+
 /// AddressFamily, in case of TCP and RDMA we use IPv6 or IPc4 only
 #[derive(Clone, Debug, Primitive)]
 pub enum AddressFamily {


nvmeadm/src/nvmf_discovery.rs
```diff
@@ -73,6 +73,19 @@ impl fmt::Display for TrType {
...
 
+impl FromStr for TrType {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "rdma" => Ok(Self::rdma),
+            "tcp"  => Ok(Self::tcp),
+            "fc"   => Ok(Self::fc),
+            _ => Err(format!("Invalid transport string for TrType")),
```
abhilashshetty04: Rust lint fails. Suggesting to use .to_string() instead of format!

abhilashshetty04 (approved):

nvmeadm/src/nvmf_discovery.rs
dsharma-dc: Modified. Linter passes now.

nvmeadm/src/nvmf_discovery.rs
```diff
@@ -73,6 +73,19 @@ impl fmt::Display for TrType {
...
 
+impl FromStr for TrType {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "rdma" => Ok(Self::rdma),
+            "tcp" => Ok(Self::tcp),
+            "fc" => Ok(Self::fc),
+            _ => Err("Invalid transport string for TrType".to_string()),
```
tiagolobocastro: might be useful to show which value was used
```suggestion
            invalid => Err(format!("Invalid transport type: {invalid}"),
```

tiagolobocastro (approved):

nvmeadm/src/nvmf_discovery.rs
dsharma-dc: yeah, displaying `invalid` now. 
On another note- since I've got my hands burned once with dependency changes merging, in this case I don't think the merging of this PR can break anything as this is just a new trait function implementation. Let me know if you think otherwise.

PR #93, opened 2024-09-11
chore: add transport to subsystem match criteria

dsharma-dc:
This shouldn't change the current behaviour. Subsequent patches will make use of this to make path disconnect decisions.

### DIFF
--- a/nvmeadm/src/nvmf_subsystem.rs
+++ b/nvmeadm/src/nvmf_subsystem.rs
@@ -1,4 +1,4 @@
-use crate::{error, parse_value};
+use crate::{error, nvmf_discovery::TrType, parse_value};
 use error::{
     nvme_error::{FileIoFailed, InvalidPath, SubsystemFailure},
     NvmeError,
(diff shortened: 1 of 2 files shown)


dsharma-dc:
This PR and https://github.com/openebs/mayastor-control-plane/pull/858 should be merged together to avoid breaking things.

nvmeadm/src/nvmf_subsystem.rs
```diff
@@ -243,15 +243,16 @@ impl Subsystem {
 
     /// Returns the particular subsystem based on the nqn and address.
     // TODO: Optimize this code.
-    pub fn get(host: &str, port: &u16, nqn: &str) -> Result<Subsystem, NvmeError> {
+    pub fn get(host: &str, port: &u16, transport: &str, nqn: &str) -> Result<Subsystem, NvmeError> {
```
tiagolobocastro: use an enum here

dsharma-dc: > use an enum here

and then convert to `&str` during comparison with `subsys.transport` ?

tiagolobocastro: Yes that's it

dsharma-dc: It's already done in latest diff, PTAL.

nvmeadm/src/nvmf_subsystem.rs
```diff
@@ -243,15 +243,23 @@ impl Subsystem {
...
-        match nvme_subsystems
-            .flatten()
-            .find(|subsys| subsys.nqn == *nqn && subsys.address.match_host_port(&host, &sport))
-        {
+        match nvme_subsystems.flatten().find(|subsys| {
+            subsys.nqn == *nqn
+                && subsys.address.match_host_port(&host, &sport)
+                && subsys
+                    .transport
+                    .eq_ignore_ascii_case(transport.to_string().as_str())
```
tiagolobocastro: You can conversion above to avoid converting to string on every iteration

nvmeadm/src/nvmf_subsystem.rs
```diff
@@ -243,15 +243,23 @@ impl Subsystem {
 
     /// Returns the particular subsystem based on the nqn and address.
     // TODO: Optimize this code.
-    pub fn get(host: &str, port: &u16, nqn: &str) -> Result<Subsystem, NvmeError> {
+    pub fn get(
```
tiagolobocastro, line 246: You can add another function, and have existing get call new function to avoid making this a breaking change

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

nvmeadm/src/nvmf_subsystem.rs
dsharma-dc: done.

nvmeadm/src/nvmf_subsystem.rs
dsharma-dc: not sure, what are the odds of this being breaking? I've changed all callers and will merge this together with control-plane changes.

avishnu (approved):

PR #96, opened 2024-10-01
feat(composer): allow setting 'host' network mode

dsharma-dc:
- Allows to set network driver as 'host' in composer
- Does not change anything in current behaviour as the current tests work without host network.
- Has limitation at the moment that container will be returning grpc handle using loopback IP, hence creating more than one container on host network isn't safe.

Can be refined further as use case evolves.

### DIFF
--- a/composer/src/composer.rs
+++ b/composer/src/composer.rs
@@ -24,6 +24,7 @@ use bollard::{
 use futures::{StreamExt, TryStreamExt};
 use ipnetwork::Ipv4Network;
 use once_cell::sync::OnceCell;
+use strum_macros::{Display, EnumString};
 
 use bollard::{
     auth::DockerCredentials,
@@ -44,6 +45,13 @@ pub const TEST_NET_NAME: &str = "engine-testing-network";
 pub const TEST_LABEL_PREFIX: &str = "io.composer.test";
 pub const TEST_NET_NETWORK: &str = "10.1.0.0/16";
 
+#[derive(Clone, EnumString, Display, Debug, PartialEq)]
+pub enum NetworkMode {
+    #[strum(serialize = "host")]
+    Host,
+    #[strum(serialize = "bridge")]
+    Bridge,
+}
 /// Additional configuration.
 #[derive(Default, Debug)]
 struct AdditionalConfig {
@@ -549,6 +557,10 @@ pub struct Builder {
     shutdown_order: Vec<ContainerName>,
     /// the network used by this experiment
     network: Ipv4Network,
+    /// network mode for the test.
+    /// XXX: Currently only use this to set 'host' mode if needed. Untested with
+    /// other modes.
+    network_mode: Option<NetworkMode>,
     /// reuse existing containers
     reuse: bool,
     /// prefix for labels set on containers and networks
@@ -595,6 +607,7 @@ impl Builder {
             existing_containers: Default::default(),
             shutdown_order: vec![],
             network: TEST_NET_NETWORK.parse().expect("Valid network config"),
+            network_mode: None,
             reuse: false,
             label_prefix: TEST_LABEL_PREFIX.to_string(),
             allow_clean_on_panic: true,
@@ -694,6 +707,12 @@ impl Builder {
         Ok(self)
     }
 
+    /// set the network mode for this test.
+    pub fn network_mode(mut self, mode: NetworkMode) -> Builder {
+        self.network_mode = Some(mode);
+        self
+    }
+
     /// the name to be used as labels and network name
     pub fn name(mut self, name: &str) -> Builder {
         self.name = name.to_string();
@@ -736,7 +755,11 @@ impl Builder {
             self.existing_containers.remove(&spec.name);
             self.containers.push((spec, next_ip));
         } else {
-            let next_ip = self.next_ip().unwrap();
+            let next_ip = if self.network_mode == Some(NetworkMode::Host) {
+                Ipv4Addr::new(127, 0, 0, 1)
+            } else {
+                self.next_ip().unwrap()
+            };
             tracing::debug!("Adding container: {}, ip: {}", spec.name, next_ip);
             self.containers.push((spec, next_ip));
         }
@@ -917,9 +940,14 @@ impl Builder {
             srcdir,
             docker,
             network_id: "".to_string(),
+            network_mode: self.network_mode.clone(),
             containers: Default::default(),
             shutdown_order: self.shutdown_order,
-            ipam,
+            ipam: if self.network_mode == Some(NetworkMode::Host) {
+                None
+            } else {
+                Some(ipam)
+            },
             label_prefix: self.label_prefix,
             reuse: self.reuse,
             allow_clean_on_panic: self.allow_clean_on_panic,
@@ -932,7 +960,12 @@ impl Builder {
             rust_log_silence: self.rust_log_silence,
         };
 
-        compose.network_id = compose.network_create().await.map_err(|e| e.to_string())?;
+        if self.network_mode == Some(NetworkMode::Host) {
+            let host_nw = &compose.host_network().await?[0];
+            compose.network_id = host_nw.id.clone().unwrap();
+        } else {
+            compose.network_id = compose.network_create().await.map_err(|e| e.to_string())?;
+        }
 
         let compose_ref = &compose;
         let create_threads = self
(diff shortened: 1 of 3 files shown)


composer/src/composer.rs
```diff
@@ -694,6 +702,12 @@ impl Builder {
         Ok(self)
     }
 
+    /// set the network mode for this test.
+    pub fn network_mode(mut self, mode: &str) -> Result<Builder, Error> {
```
tiagolobocastro: What is mode? Suggest you use an enum for the public facing api

composer/src/composer.rs
```diff
@@ -932,7 +949,12 @@ impl Builder {
             rust_log_silence: self.rust_log_silence,
         };
 
-        compose.network_id = compose.network_create().await.map_err(|e| e.to_string())?;
+        if self.network_mode.is_none() {
+            compose.network_id = compose.network_create().await.map_err(|e| e.to_string())?;
+        } else {
+            let host_nw = &compose.host_network().await?[0];
```
tiagolobocastro, line 964: what is [0] index here?

composer/src/composer.rs
```diff
@@ -932,7 +949,12 @@ impl Builder {
...
+            compose.network_id = host_nw.id.clone().unwrap();
```
tiagolobocastro, line 965: Should return an error if no [0] and this unwrap

composer/src/composer.rs
```diff
@@ -1282,7 +1319,7 @@ impl ComposeTest {
     async fn create_container(
         &self,
         spec: &ContainerSpec,
-        ipv4: Ipv4Addr,
+        ipv4: Option<Ipv4Addr>,
```
tiagolobocastro: Why is the ip optional, because it's not in control of docker so it's up to the host?
Can we not set 127.0.0.1 in case of host?

composer/src/composer.rs
```diff
@@ -1368,16 +1406,16 @@ impl ComposeTest {
...
                 }),
                 aliases: spec.network_aliases.clone(),
                 ..Default::default()
             },
         );
 
-        let mut env = self.spec_environment(spec);
-        env.push(format!("MY_POD_IP={ipv4}"));
+        let mut env = spec.environment();
+        env.push(format!("MY_POD_IP={:?}", ipv4.unwrap_or(Ipv4Addr::new(127, 0, 0, 1))));
```
tiagolobocastro: Ah so every container gets the same ip?

composer/src/composer.rs
dsharma-dc: This maps to network mode of `HostConfig` in bollard crate. I'll add an enum for our test facing API.
```
    /// Network mode to use for this container. Supported standard values are: `bridge`, `host`, `none`, and `container:<name|id>`. Any other value is taken as a custom network's name to which this container should connect to. 
    #[serde(rename = "NetworkMode")]
    #[serde(skip_serializing_if = "Option::is_none")]
    pub network_mode: Option<String>,
```

composer/src/composer.rs
dsharma-dc: `host_network()` uses the `list_networks()` API from library since there is no specific one to fetch a single network. We are filtering by driver `host`, so what we get is a list of `host` networks which will realistically be always a single host network on a system. Just taking first entry out of it using [0].

composer/src/composer.rs
dsharma-dc: There should always be a host network on the node, so unwrapping panic should be a fair stopper here?

composer/src/composer.rs
dsharma-dc: This IP gets sent down to docker API to create container, and if it has a value set along with `host` network mode, the API fails complaining that with `host` network IP can't be provided by caller.

composer/src/composer.rs
dsharma-dc: Actually, setting this env for host mode io-engine containers isn't of any relevance here because we'll be providing interface specifically via `-T` arg, but yeah it makes all of them see same MY_POD_IP.

However, in a follow up patch (https://github.com/openebs/mayastor/pull/1759) we make the io-engine-test API return 127.0.0.1 as grpc endpoint IP for all such containers.

composer/src/composer.rs
tiagolobocastro: Alright seems reasonable
( jk on the other PR I had the Some was also always guaranteed, yet you asked me to handle it there :))

composer/src/composer.rs
tiagolobocastro: I see we already do unwrap on ids else in the code, so it's fine I guess

composer/src/composer.rs
tiagolobocastro: I see, makes sense

composer/src/composer.rs
tiagolobocastro: hmm can we leave the containers as ip and then just unwrap to 127.0.0.1 here itself to avoid changing the public api?

composer/src/composer.rs
dsharma-dc: 8-) 
more liberal to use unwraps and expects in test code. composer is a middle ground, but still test utility :)

composer/src/composer.rs
dsharma-dc: > hmm can we leave the containers as ip and then just unwrap to 127.0.0.1 here itself to avoid changing the public api?

Are you suggesting:
1. set the container's IP from here in hashmap of (containerId, ip address) as 127.0.0.1, and
2. let the `grpc_handle()` API internals be as it is today and expect the ip available always while returning SocketAddr?

```
format!("{}:10124", v.1 .1.unwrap_or(Ipv4Addr::new(127, 0, 0, 1)))
                        .parse::<std::net::SocketAddr>()
                        .unwrap()
```

tiagolobocastro: Yeah, basically just make sure the ip is set on the container hashmap

dsharma-dc: > Yeah, basically just make sure the ip is set on the container hashmap

I think I had tried that and ran into some issue. Since I don't remember what trouble I faced, I'll try once again.

dsharma-dc: I've simplified now, keeping hashmap as-is. Setting ip as localhost early on, but handling based on network mode just before calling docker create_container.

composer/src/composer.rs
```diff
@@ -735,6 +754,8 @@ impl Builder {
             let next_ip = container.1;
             self.existing_containers.remove(&spec.name);
             self.containers.push((spec, next_ip));
+        } else if self.network_mode == Some(NetworkMode::Host) {
+            self.containers.push((spec, Ipv4Addr::new(127, 0, 0, 1)));
         } else {
             let next_ip = self.next_ip().unwrap();
```
tiagolobocastro: Let's keep logging

```suggestion
        } else {
            let next_ip = if self.network_mode == Some(NetworkMode::Host) {
                Ipv4Addr::new(127, 0, 0, 1)))
            } else {
                self.next_ip().unwrap()
            };
```

composer/src/composer.rs
```diff
@@ -1573,7 +1622,7 @@ impl ComposeTest {
                 ),
             })?;
         if !self.reuse || self.prune_matching {
-            tracing::debug!("Starting container: {}, ip: {}", name, id.1);
+            tracing::debug!("Starting container: {}, ip: {:?}", name, id.1);
```
tiagolobocastro: This can be reverted?

```suggestion
            tracing::debug!("Starting container: {}, ip: {}", name, id.1);
```

composer/src/composer.rs
```diff
@@ -549,6 +557,10 @@ pub struct Builder {
     shutdown_order: Vec<ContainerName>,
     /// the network used by this experiment
     network: Ipv4Network,
+    /// network mode for the test.
```
tiagolobocastro, line 560: Ah we had network here already, we couldn't make use of that as host?

tiagolobocastro (approved):

composer/src/composer.rs
dsharma-dc: I noticed it, but I don't think it's of any use for host network mode. To use it we have to set ip range in it and then that's used to configure IPAM, we don't need all those.

Abhinandan-Purkait (approved):

composer/src/composer.rs
```diff
@@ -694,6 +707,12 @@ impl Builder {
         Ok(self)
     }
 
+    /// set the network mode for this test.
+    pub fn network_mode(mut self, mode: NetworkMode) -> Result<Builder, Error> {
```
tiagolobocastro: Just noticed this, why the Error?

dsharma-dc: eh, good sighting. It came with copying from existing functions. Will return just `Builder`

dsharma-dc: fixed now.

PR #98, opened 2024-10-08
build: upgrade nix-shell helm to v3.15.4

niladrih:
Helm's upgrade option --reset-then-reuse-values requires helm 3.14 or above. Using Helm v3.15.4 across the Mayastor repos to support use for this option in the upgrade-job and/or E2e tests.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -399,7 +399,7 @@ exec_helm() {
   if [ -n "$LOCAL_HELM" ]; then
     $HELM $@
   else
-    $NIX_SHELL -p "(import (import $NIX_SOURCES).nixpkgs {}).kubernetes-helm-wrapped" --run "$HELM $@"
+    $NIX_SHELL -p "(import (import $NIX_SOURCES).helm-nixpkgs {}).kubernetes-helm" --run "$HELM $@"
   fi
 }
 
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -1,4 +1,16 @@
 {
+    "helm-nixpkgs": {
+        "branch": "master",
+        "description": "Nix Packages collection & NixOS",
+        "homepage": "",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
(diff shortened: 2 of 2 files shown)


niladrih:
PR 553 on mayastor-extensions requires this.

shell.nix
```diff
@@ -19,6 +19,7 @@ pkgs.mkShell {
     cacert
     cargo-udeps
     clang
+    niv
```
tiagolobocastro: nit: might suggest leave niv out, you don't need to enter this nix-shell, simple `nix-shell -p niv`

scripts/release.sh
```diff
@@ -399,7 +399,7 @@ exec_helm() {
   if [ -n "$LOCAL_HELM" ]; then
     $HELM $@
   else
-    $NIX_SHELL -p "(import (import $NIX_SOURCES).nixpkgs {}).kubernetes-helm-wrapped" --run "$HELM $@"
+    $NIX_SHELL -p "(import (import $NIX_SOURCES).helm-nixpkgs {}).kubernetes-helm" --run "$HELM $@"
```
tiagolobocastro, line 402: Can't we just update the nixpkgs?

niladrih: That'd get us to 3.15.0, I was leaning more towards 3.15.4. WDYT?

shell.nix
niladrih: Sure..

scripts/release.sh
niladrih: Alright, I guess 3.15.0 is just as good as 3.15.4. Closing this PR.

niladrih:
This is not needed. Upgrading the nixpkgs 24.05 release branch's revision should pull in helm 3.15.0, which is just as good as 3.15.4.

scripts/release.sh
tiagolobocastro: But we still need to upgrade nixpkgs, or do we already have 3.15.0 here?

PR #100, opened 2024-10-22
build: add support for extra nix-build args

niladrih:
Add support for providing custom nix-build args to the release.sh.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -261,6 +261,11 @@ parse_common_arg() {
       STATIC_LINKING="false"
       shift
       ;;
+    --extra-build-args)
+      shift
+      CUSTOM_BUILD_ARGS="$1"
+      shift
+      ;;
     --skip-bins)
       shift
       BUILD_BINARIES=
@@ -334,6 +339,9 @@ setup() {
     NIX_BUILD="$NIX_BUILD $NIX_TAG_ARGS"
     TAG="$ALIAS_TAG"
     ALIAS_TAG=
+  fi
+  if [ -n "$CUSTOM_BUILD_ARGS" ]; then
+    NIX_BUILD="$NIX_BUILD $CUSTOM_BUILD_ARGS"
   fi
 }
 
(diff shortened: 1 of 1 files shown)


Abhinandan-Purkait (approved):

scripts/release.sh
```diff
@@ -261,6 +261,10 @@ parse_common_arg() {
       STATIC_LINKING="false"
       shift
       ;;
+    --no-debug-symbols)
```
tiagolobocastro: suggestion: strip symbols is a well know thing

```suggestion
    --strip)
```

scripts/release.sh
```diff
@@ -459,7 +463,7 @@ build_images() {
     # the images we already have locally.
     if [ -z "$SKIP_BUILD" ]; then
       echo "Building $image:$TAG ..."
-      $NIX_BUILD --out-link "$archive-image" -A "images.$BUILD_TYPE.$archive" --arg allInOne "$ALL_IN_ONE" --arg incremental "$INCREMENTAL" --argstr product_prefix "$PRODUCT_PREFIX"
+      $NIX_BUILD --out-link "$archive-image" -A "images.$BUILD_TYPE.$archive" --arg allInOne "$ALL_IN_ONE" --arg noDebugSymbols "$DISABLE_DEBUG_SYMBOLS" --arg incremental "$INCREMENTAL" --argstr product_prefix "$PRODUCT_PREFIX"
```
tiagolobocastro: Just wondering, does it make sense to also allow passing custom builds flags?

tiagolobocastro (approved):

scripts/release.sh
niladrih: Yes that would be easier. I'll rework this PR to do that.

scripts/release.sh
niladrih: strip isn't very clear on what this will eventually do?
`-C strip=debuginfo -C debuginfo=0` would avoid generating debug symbols as well. Also it wouldn't remove the entire symbols table, just the debug symbols.

Anyhow, this PR's new direction is to allow random build flags, so this might not matter all that much.

dsharma-dc: On a related note - Have you considered splitting the debuginfo and symbols to a separate file using `split-debuginfo` , to reduce the size while still keeping potentially useful info around?

scripts/release.sh
niladrih: I've made this change, PTAL.

scripts/release.sh
niladrih: This is no longer relevant, I think.

niladrih: Going with a generic nix-build args input to the release.sh script. This opens it up for custom build parameters. via a custom nix variable which I'll add eventually in separate PRs to the code repos.

PR #107, opened 2025-01-03
chore: elaborate nvmf subsystem error

dsharma-dc:


nvmeadm/src/nvmf_discovery.rs
```diff
@@ -61,6 +61,18 @@ pub enum TrType {
     tcp = 3,
 }
 
+impl TrType {
+    // Method to convert an i32 to TrType
+    pub fn from_i32(value: i32) -> Option<Self> {
```
tiagolobocastro: Isn't this auto generated by Primitive derive?

dsharma-dc: ok, It works using `TrType::try_from()` as we already have `enum_primitive_derive` dependency. Will remove this impl.

dsharma-dc: done

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

PR #110, opened 2025-02-06
feat(pool/encryption): add api messages

dsharma-dc:
Add new API messages for pool encryption support.

### DIFF
--- a/apis/io-engine/protobuf/v1/common.proto
+++ b/apis/io-engine/protobuf/v1/common.proto
@@ -23,3 +23,31 @@ message MayastorBugFixes {
   // Nexus rebuilds both the clusters allocated to the replica and its ancestors clusters
   bool nexusRebuildReplicaAncestry = 1;
 }
+
+// Cipher to use for encryption.
+enum Cipher {
+  AES_CBC = 0;
+  AES_XTS = 1;
+}
+
+// Represents an encryption key that can be used to encrypt an
+// entity like pool or lvol/replica.
+message EncryptionKey {
+  // Name of the key.
+  string key_name = 1;
+  // The AES encryption key.
+  bytes key = 2;
+  // AES Key length.
+  uint32 key_length = 3;
+  // key2 (required for AES_XTS).
+  optional bytes key2 = 4;
+  // The length of key2. Must be same as key_length.
+  optional uint32 key2_length = 5;
+}
+
+message Encryption {
+  // Cipher to be used.
+  Cipher cipher = 1;
+  // The encryption key.
+  EncryptionKey key = 2;
+}
--- a/apis/io-engine/protobuf/v1/pool.proto
+++ b/apis/io-engine/protobuf/v1/pool.proto
@@ -3,6 +3,7 @@ syntax = "proto3";
 
 import "google/protobuf/empty.proto";
 import "google/protobuf/wrappers.proto";
+import "common.proto";
 
 package mayastor.v1;
 
@@ -48,6 +49,8 @@ message CreatePoolRequest {
   optional uint32 cluster_size = 5;
   // Optional pool metadata arguments.
   optional PoolMetadataArgs md_args = 6;
+  // Encryption parameters for this pool.
+  optional Encryption encryption = 7;
 }
 
 // Pool metadata arguments.
@@ -77,6 +80,8 @@ message ImportPoolRequest {
   google.protobuf.StringValue uuid = 2; // optional uuid for the pool to be imported
   repeated string disks = 3; // disk device paths or URIs to be claimed by the pool
   PoolType pooltype = 4;     // type of the pool
+  // Encryption parameters for this pool.
+  optional Encryption encryption = 5;
 }
 
 // State of the storage pool (terminology comes from ZFS).
--- a/apis/io-engine/src/v1.rs
+++ b/apis/io-engine/src/v1.rs
@@ -11,7 +11,7 @@ mod pb {
 }
 
 pub mod common {
-    pub use super::pb::ShareProtocol;
+    pub use super::pb::{Cipher, Encryption, EncryptionKey, ShareProtocol};
 }
 
 /// v1 version of bdev grpc API


dsharma-dc:
DO_NOT_MERGE: till the dependent code scaffolding is available in core-agent and io-engine.

apis/io-engine/protobuf/v1/common.proto
```diff
@@ -23,3 +23,31 @@ message MayastorBugFixes {
...
+  // The length of key2. Must be same as key_length.
+  optional uint32 key2_length = 5;
+}
+
+message EncryptionInfo {
+  // Cipher to be used.
+  Cipher cipher = 1;
+  // The encryption key.
+  EncryptionKey key = 2;
+}
```
Abhinandan-Purkait: Blank line needed at the end?

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -77,6 +80,8 @@ message ImportPoolRequest {
   google.protobuf.StringValue uuid = 2; // optional uuid for the pool to be imported
   repeated string disks = 3; // disk device paths or URIs to be claimed by the pool
   PoolType pooltype = 4;     // type of the pool
+  // Encryption key for this pool.
```
Abhinandan-Purkait:
```suggestion
  // Encryption parameters for this pool.
```

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -48,6 +49,8 @@ message CreatePoolRequest {
   optional uint32 cluster_size = 5;
   // Optional pool metadata arguments.
   optional PoolMetadataArgs md_args = 6;
+  // Encryption key for this pool.
```
Abhinandan-Purkait:
```suggestion
  // Encryption parameters for this pool.
```

Abhinandan-Purkait (approved):

apis/io-engine/protobuf/v1/common.proto
```diff
@@ -23,3 +23,31 @@ message MayastorBugFixes {
...
+  AES_CBC = 0;
+  AES_XTS = 1;
+}
+
+// Represents an encryption key that can be used to encrypt an
+// entity like pool or lvol/replica.
+message EncryptionKey {
+  // Name of the key.
+  string key_name = 1;
+  // The AES encryption key.
```
tiagolobocastro, line 38: Is this specific for AES?
If so, should this be an AES specific EncryptionKey message?

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -48,6 +49,8 @@ message CreatePoolRequest {
...
+  optional EncryptionInfo enc_info = 7;
```
```suggestion
  optional EncryptionInfo encryption = 7;
```

apis/io-engine/protobuf/v1/common.proto
```diff
@@ -23,3 +23,31 @@ message MayastorBugFixes {
...
+  bytes key = 2;
+  // AES Key length.
+  uint32 key_length = 3;
+  // key2 (required for AES_XTS).
+  optional bytes key2 = 4;
+  // The length of key2. Must be same as key_length.
+  optional uint32 key2_length = 5;
+}
+
+message EncryptionInfo {
```
tiagolobocastro:
```suggestion
message Encryption {
```

apis/io-engine/protobuf/v1/common.proto
```diff
@@ -23,3 +23,31 @@ message MayastorBugFixes {
   // Nexus rebuilds both the clusters allocated to the replica and its ancestors clusters
   bool nexusRebuildReplicaAncestry = 1;
 }
+
+// Cipher to use for encryption.
+enum Cipher {
```
tiagolobocastro, line 28: nit: because others have Encryption as prefix?

```suggestion
enum EncryptonCipher {
```

tiagolobocastro (commented):
OEP needs to be approved before merging this?

apis/io-engine/protobuf/v1/common.proto
dsharma-dc: > Is this specific for AES? If so, should this be an AES specific EncryptionKey message?

AES is the only algo we and spdk support for foreseeable future, as that's the industry standard. Hence this is kept in accordance with AES ciphers. Only XTS ciphermode needs two keys, so if any other mode is supported it can accommodate in this message.

apis/io-engine/protobuf/v1/common.proto
dsharma-dc: > nit: because others have Encryption as prefix?

Though `Cipher` seems to be the de-facto and more widely used term in cryptography to refer to a cipher mode.

tiagolobocastro: Fair enough

dsharma-dc:
> OEP needs to be approved before merging this?

OEP is now set as `implementable`.

apis/io-engine/protobuf/v1/common.proto
```diff
@@ -23,3 +23,31 @@ message MayastorBugFixes {
...
+
+// Represents an encryption key that can be used to encrypt an
+// entity like pool or lvol/replica.
+message EncryptionKey {
+  // Name of the key.
+  string key_name = 1;
+  // The AES encryption key.
+  bytes key = 2;
+  // AES Key length.
+  uint32 key_length = 3;
```
tiagolobocastro, line 41: Is the key len needed? Doesn't bytes already contain the len?

dsharma-dc: It's generally a practice and a direct way of telling whether we're using 128-bit key or 256-bit key. Also, good to have for some validations we might do "Is the key bytes really as long as the cipher needs and asks".

tiagolobocastro: SPDK seems to just pass the bytes: <https://github.com/spdk/spdk/blob/e01cb43b8578f9155d07a9bc6eee4e70a3af96b0/proto/sma.proto#L27>
I think bytes would be enough, but len is also fine for validation purposes

tiagolobocastro (approved):

PR #112, opened 2025-02-25
feat(encryption): adds the crate to fetch the secret

Abhinandan-Purkait:
- Adds a crate to fetch a secret either via file or k8s secret.

### DIFF
--- a/secret_provider/Cargo.toml
+++ b/secret_provider/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "secret-provider"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+async-trait = "0.1"
+tokio = { version = "1", features = ["fs", "macros", "rt-multi-thread"] }
+k8s-openapi = { version = "0.22.0", features = ["v1_24"] }
+kube = { version = "0.94.2", features = ["derive", "runtime"] }
+snafu = "0.8.5"
--- a/secret_provider/src/error.rs
+++ b/secret_provider/src/error.rs
@@ -1,0 +1,24 @@
+#[derive(Debug, snafu::Snafu)]
+#[snafu(visibility(pub), context(suffix(false)))]
+pub enum Error {
+    #[snafu(display("Failed to read file '{file}': {source}"))]
+    ReadFile {
+        source: std::io::Error,
+        file: String,
+    },
+
+    #[snafu(display("Failed to parse JSON: {source}"))]
+    ParseJson { source: serde_json::Error },
+
+    #[snafu(display("Kubernetes API error: {source}"))]
+    KubeApi { source: kube::Error },
+
+    #[snafu(display("Kubernetes secret is missing data"))]
+    MissingSecretData {},
+
+    #[snafu(display("Key '{key}' not found in secret data"))]
+    MissingSecretKey { key: String },
+
+    #[snafu(display("Failed to convert secret data to UTF-8: {source}"))]
+    Utf8Conversion { source: std::str::Utf8Error },
+}
--- a/secret_provider/src/file.rs
+++ b/secret_provider/src/file.rs
@@ -1,0 +1,36 @@
+use crate::{error::Error, SecretProvider};
+use std::path::PathBuf;
+
+use async_trait::async_trait;
+use serde::de::DeserializeOwned;
+
+/// A secret provider that reads a file from the provided path.
+pub struct FileSecretProvider {
+    base_path: PathBuf,
+}
+
+impl FileSecretProvider {
+    /// Create a new `FileSecretProvider` using the provided base path.
+    pub fn new(base_path: PathBuf) -> Self {
+        Self { base_path }
+    }
+}
+
+#[async_trait]
+impl SecretProvider for FileSecretProvider {
+    async fn secret_data<T>(&self, identifier: &str) -> Result<T, Error>
+    where
+        T: DeserializeOwned + Send,
+    {
+        let file_path = self.base_path.join(identifier);
+        let content = tokio::fs::read_to_string(file_path)
+            .await
+            .map_err(|error| Error::ReadFile {
+                source: error,
+                file: identifier.to_string(),
+            })?;
+        let secret =
+            serde_json::from_str(&content).map_err(|error| Error::ParseJson { source: error })?;
+        Ok(secret)
+    }
+}
--- a/secret_provider/src/k8s.rs
+++ b/secret_provider/src/k8s.rs
@@ -1,0 +1,78 @@
+//! This module reads from a secret provided the given key matches. Consider the example as below.
+//!
+//!```yaml
+//! apiVersion: v1
+//! kind: Secret
+//! metadata:
+//!   name: my-secret
+//!   namespace: default
+//! type: Opaque
+//! stringData:
+//!   encryption_parameters: |
+//!    {
+//!       "cipher": "AesCbc",
+//!       "key": {
+//!        "key_name": "example-key",
+//!        "key": "2b7e151628aed2a6abf7158809cf4f3c",
+//!         "key_length": 128
+//!       }
+//!     }
+//! ```
+//!
+use crate::SecretProvider;
+
+use crate::error::Error;
+use async_trait::async_trait;
+use k8s_openapi::api::core::v1::Secret;
+use serde::de::DeserializeOwned;
+
+const SECRET_PARAMS_KEY: &str = "encryption_parameters";
+
+/// A secret provider that retrieves a Kubernetes secret.
+pub struct K8sSecretProvider {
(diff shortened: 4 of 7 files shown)


secret_provider/src/k8s.rs
```diff
@@ -0,0 +1,82 @@
...
+//!   name: my-secret
+//!   namespace: default
+//! type: Opaque
+//! stringData:
+//!   encryption_parameters: |
+//!    {
+//!       "cipher": "AesCbc",
+//!       "key": {
+//!        "key_name": "example-key",
+//!        "key": [115, 111, 109, 101, 45, 107, 101, 121],
```
Abhinandan-Purkait: Reviewers: Should this be plain text?

secret_provider/src/types.rs
```diff
@@ -0,0 +1,34 @@
+use serde::{Deserialize, Serialize};
+
+/// Encryption parameters.
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
```
Abhinandan-Purkait: Reviewers: We should use this type by wrapping it in a newtype in the control-plane and data-plane, rather than duplicating. Is that a fair idea?

secret_provider/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "secret_provider"
```
tiagolobocastro:
```suggestion
name = "secret-provider"
```

secret_provider/src/error.rs
```diff
@@ -0,0 +1,24 @@
+#[derive(Debug, snafu::Snafu)]
+#[snafu(visibility(pub), context(suffix(false)))]
+pub enum Error {
+    #[snafu(display("Failed to read file '{}': {}", file, source))]
```
tiagolobocastro: Do this for all

```suggestion
    #[snafu(display("Failed to read file '{file}': {source}", file, source))]
```

secret_provider/src/k8s.rs
tiagolobocastro: Typically this would be md5sum kind of thing here

secret_provider/src/k8s.rs
```diff
@@ -0,0 +1,82 @@
...
+use k8s_openapi::api::core::v1::Secret;
+
+const SECRET_DATA_KEY: &str = "encryption_parameters";
+
+/// A secret provider that retrieves a Kubernetes secret.
+pub struct K8sSecretProvider {
+    /// The Kubernetes client.
+    pub client: kube::Client,
+    /// The namespace in which to look for the secret.
+    pub namespace: String,
```
tiagolobocastro: You could take it from the client default namespace?

secret_provider/src/k8s.rs
```diff
@@ -0,0 +1,82 @@
...
+            .map_err(|e| Error::KubeApi { source: e })?;
+
+        let data = secret.data.ok_or_else(|| Error::MissingSecretData {})?;
+        let raw_bytes = data
+            .get(&self.secret_data_key)
+            .ok_or_else(|| Error::MissingSecretKey {
+                key: self.secret_data_key.to_string(),
+            })?
+            .0
+            .clone();
```
tiagolobocastro: don't need to clone? The reference is used below

secret_provider/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+use types::EncryptionParameters;
+
+use async_trait::async_trait;
+
+/// A trait for loading encryption parameters from a secret source.
+///
+/// Implementors can load from various backends (file, Kubernetes API, etc.).
+#[async_trait]
+pub trait SecretProvider {
+    async fn get_encryption_parameters(
```
tiagolobocastro:
```suggestion
    async fn encryption_parameters(
```

secret_provider/src/types.rs
tiagolobocastro: hmm how about decoupling the types from the `SecretProvider`? 
So basically `SecretProvider` will only know how to return a given type which implements deserialize, and then another abstraction can convert to the type we want.
Then we can have these types be the default, but also makes it easy to use a different type should we need to.

secret_provider/src/k8s.rs
Abhinandan-Purkait: Yeah, I meant should we make it string rather than u8 array

secret_provider/src/k8s.rs
Abhinandan-Purkait: That might not always work, if it's a non mayastor namespace

Abhinandan-Purkait: It's supposed to be passed down by the service that is using this crate, as it would in the same ns.

secret_provider/src/k8s.rs
Abhinandan-Purkait: Doesn't work without the clone.

secret_provider/src/types.rs
Abhinandan-Purkait: Done.

secret_provider/src/k8s.rs
```diff
@@ -0,0 +1,79 @@
...
+//!     }
+//!
+use crate::SecretProvider;
+
+use crate::error::Error;
+use async_trait::async_trait;
+use k8s_openapi::api::core::v1::Secret;
+use serde::de::DeserializeOwned;
+
+const SECRET_DATA_KEY: &str = "encryption_parameters";
```
dsharma-dc: Shall we call this SECRET_PARAMS_KEY. `DATA_KEY` confuses here with the actual DEK that's `encryption_parameters.key`.

secret_provider/src/k8s.rs
```diff
@@ -0,0 +1,79 @@
...
+//! stringData:
+//!   encryption_parameters: |
+//!    {
+//!       "cipher": "AesCbc",
+//!       "key": {
+//!        "key_name": "example-key",
+//!        "key": "2b7e151628aed2a6abf7158809cf4f3c",
+//!         "key_length": 128,
+//!         "key2": null,
+//!         "key2_length": null
```
dsharma-dc: You can fix this example - omit the key2 and key2_length for AesCbc. key2 is only relevant for AesXts. We'll add those validations in this crate?

secret_provider/src/k8s.rs
dsharma-dc: Yes, a string in hex representation.

secret_provider/src/file.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use async_trait::async_trait;
+use serde::de::DeserializeOwned;
+
+/// A secret provider that reads a file from the provided path.
+pub struct FileSecretProvider;
+
+#[async_trait]
+impl SecretProvider for FileSecretProvider {
+    async fn encryption_parameters<T>(&self, file_path: &str) -> Result<T, Error>
```
tiagolobocastro: Since we decoupled the encryption from the secret

```suggestion
    async fn secret_data<T>(&self, file_path: &str) -> Result<T, Error>
```

tiagolobocastro: And also, should be `identifier` rather than `file_path`.
File path we need to figure out based on self.base_path or something

secret_provider/src/k8s.rs
tiagolobocastro: Can you clarify "doesn't work"?

secret_provider/src/k8s.rs
Abhinandan-Purkait: Done.

secret_provider/src/file.rs
Abhinandan-Purkait: Done.

secret_provider/src/k8s.rs
Abhinandan-Purkait: Done.

dsharma-dc (approved):

secret_provider/src/file.rs
tiagolobocastro, line 1: We're missing a new for the FileSecretProvier with a base path for the file?

secret_provider/src/file.rs
```diff
@@ -0,0 +1,25 @@
...
+/// A secret provider that reads a file from the provided path.
+pub struct FileSecretProvider;
+
+#[async_trait]
+impl SecretProvider for FileSecretProvider {
+    async fn secret_data<T>(&self, identifier: &str) -> Result<T, Error>
+    where
+        T: DeserializeOwned + Send,
+    {
+        let content = tokio::fs::read_to_string(identifier)
```
tiagolobocastro: Should be base path + identifier ?

tiagolobocastro (approved):

secret_provider/src/file.rs
Abhinandan-Purkait: Done.

secret_provider/src/file.rs
Abhinandan-Purkait: Done.

PR #117, opened 2025-03-11
chore: rename crate to secret-provider

dsharma-dc:


secret-provider/examples/src/main.rs
```diff
@@ -44,7 +44,7 @@ async fn main() -> Result<(), Error> {
     println!("Encryption Parameters from k8s: {:?}", raw_secret);
 
     let k8s_provider =
-        FileSecretProvider::new("/root/work/openebs/mayastor-dependencies/secret_provider".into());
+        FileSecretProvider::new("/root/work/openebs/mayastor-dependencies/secret-provider".into());
```
tiagolobocastro: nit: should this be a file checked in to examples? Otherwise this can't be run by default

tiagolobocastro (approved):

secret-provider/examples/src/main.rs
dsharma-dc: > nit: should this be a file checked in to examples? Otherwise this can't be run by default

I've added a sample file now and made pathbuf generic to use pwd. It's going to fail for k8s secret provider but works for file provider.

secret-provider/examples/Cargo.toml
```diff
@@ -4,11 +4,13 @@ version = "0.1.0"
 edition = "2021"
 
 [[bin]]
-name = "secret_provider"
+name = "secret-provider-example"
```
tiagolobocastro: example should be [[example]] rather than bin

abhilashshetty04 (approved):

PR #119, opened 2025-03-18
chore: expose helper utility to fetch secret from source

dsharma-dc:


secret-provider/src/lib.rs
```diff
@@ -3,10 +3,53 @@ pub mod file;
...
 
 use error::Error;
+use tracing::trace;
 
 use async_trait::async_trait;
+use platform::{current_platform_type, PlatformType};
 use serde::de::DeserializeOwned;
 
+/// Parse and return encryption parameters from a Kubernetes secret or a file.
+pub async fn params_get_from_secret_source<T: DeserializeOwned + Send>(
```
tiagolobocastro:
```suggestion
pub async fn secret_data<T: DeserializeOwned + Send>(
```

secret-provider/src/lib.rs
```diff
@@ -3,10 +3,53 @@ pub mod file;
...
+    secret_source_name: &str,
+) -> Result<T, Error> {
+    let secret_params = match current_platform_type() {
+        PlatformType::K8s => {
+            let client = kube::Client::try_default()
+                .await
+                .map_err(|e| Error::KubeApi { source: e })?;
+            // todo: Add cli arg for namespace as well?
+            let ns = client.default_namespace().to_string();
```
tiagolobocastro: If running from cli I guess we could read something like env `MY_POD_NAMESPACE` ?

```suggestion
            let ns = match std::env::var("MY_POD_NAMESPACE") {
              Ok(ns) if !ns.is_empty() => ns,
              _ => client.default_namespace().to_string()
            };
```

secret-provider/src/lib.rs
```diff
@@ -3,10 +3,53 @@ pub mod file;
...
+            trace!(
+                "Platform: Deployer. read secret params from file {:?}",
+                secret_source_name
+            );
+
+            let secret_params: Option<T> = file_provider.secret_data(secret_source_name).await?;
+
+            secret_params
+        }
+        PlatformType::None => unreachable!(),
```
tiagolobocastro: Should we also read the file, or at least return None, otherwise this will crash when running binary directly?

dsharma-dc: > Should we also read the file, or at least return None, otherwise this will crash when running binary directly?

No we don't crash running io-engine binary directly today as it takes the Encryption::Data() path which doesn't call library, and io-engine-client has only that provision.
If you think io-engine-client should have provision to take filename, then I can create a task and then change this match branch to read file also.

secret-provider/src/lib.rs
dsharma-dc: > If running from cli I guess we could read something like env `MY_POD_NAMESPACE` ?

Will that be relevant here though if we are running outside of k8s cluster? we'll not take this route.
Or you're saying from a perspective if we want to for example edit daemonset and pass this env?

dsharma-dc: MY_POD_NAMESPACE looks like a k8s standard env, so might conflict with the correct (`client.default_namespace()`) namespace as well?

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
cargo fmt needed it seems?

secret-provider/src/lib.rs
tiagolobocastro: I mean, if you specify a secret name on the io-engine-client?

secret-provider/src/lib.rs
tiagolobocastro: io-engine can be run outside of a k8s pod as long as the connections are setup correctly
Up till now, io-engine had no k8s connection, so this namespace was not required, but in this case we have to specify correct namespace if we want it to find the secrets.
If you have a look at platform, there's already uses of `MY_POD_NAMESPACE`.

secret-provider/src/lib.rs
dsharma-dc: > I mean, if you specify a secret name on the io-engine-client?

Currently I haven't provided a cli option to take secret name from `io-engine-client`. It just takes raw params.

tiagolobocastro: Ah I see
Still I'd suggest returning an error here, given this is a library and may be called from other places eventually (ex: core-agent, etc)

dsharma-dc: done

secret-provider/src/lib.rs
dsharma-dc: done

dsharma-dc:
> cargo fmt needed it seems?

ran it on secret-provider crate now.

tiagolobocastro (approved):

PR #120, opened 2025-03-21
fix(nvme): remove [ ] from the host

tiagolobocastro:
The Url crate keeps the [ ] when it returns the host so we need to remove it ourselves.


tiagolobocastro:
Relates to https://github.com/openebs/mayastor/issues/1731

Abhinandan-Purkait (approved):

dsharma-dc (approved):

PR #123, opened 2025-04-30
build: add support for the openebs/openebs helm chart during helm_build_deps

niladrih:


scripts/release.sh
```diff
@@ -423,9 +423,17 @@ build_helm_deps() {
   fi
 
   echo "Updating helm chart dependencies ..."
-  # Helm chart directory path -- /scripts --> /chart
-  chart_dir="$SCRIPT_DIR/../chart"
-  dep_chart_dir="$chart_dir/charts"
+
+  # Helm chart directory path -- /scripts --> /chart (or /charts)
+  local chart_dir=
```
tiagolobocastro: Let's instead specify this, like we do for IMAGES:
```
IMAGES="agents.core agents.ha.node agents.ha.cluster operators.diskpool rest csi.controller csi.node"
```
and do like
```
CHART=.....
```

niladrih: Yes, this was a stupid way to handle this. 👍

niladrih: Done. PTAL.

scripts/release.sh
```diff
@@ -423,9 +423,9 @@ build_helm_deps() {
   fi
 
   echo "Updating helm chart dependencies ..."
-  # Helm chart directory path -- /scripts --> /chart
-  chart_dir="$SCRIPT_DIR/../chart"
-  dep_chart_dir="$chart_dir/charts"
+
+  # Helm chart directory path -- /scripts --> /chart (or /charts)
+  local dep_chart_dir="$HELM_CHART_DIR/charts"
```
tiagolobocastro: Maybe also skip this on error out if `build_deps` is set but `HELM_CHART_DIR` is not set?

tiagolobocastro (approved):

dsharma-dc (approved):

Abhinandan-Purkait (approved):

PR #129, opened 2025-07-14
build: remove step to unpack helm chart tarballs

niladrih:
The mayastor upgrade-job reads through helm dependency chart values. The
build step which performs the helm dependency update also unpacks the
dependency chart tarballs.

Helm works in unexpected ways if there exists a tarball and a directory
for the same dependency chart, both in the charts directory.

The mayastor upgrade-job could unpack the tarballs in its own nix build
sandbox. Alternatively, it could use the helm show values command.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,6 @@ Cargo.lock
 /package.json
 /rust-git-version
 /rust-toolchain.toml
-.idea
+**/.idea
 .vscode
 .pregenerated
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -431,15 +431,10 @@ build_helm_deps() {
   local chart_dir=${HELM_CHART_DIR%/}
   local dep_chart_dir="$chart_dir/charts"
 
-  # This performs a dependency update and then extracts the tarballs pulled.
-  # If and when the `--untar` functionality is added to the `helm dependency
-  # update command, the for block can be removed in favour of the `--untar` option.
-  # Ref: https://github.com/helm/helm/issues/8479
+  # Helm dependency update doesn't work correctly if a version of a dependency is newer than the required version
+  # This rm command clears out existing chart tarballs, so that helm dependency update works correctly.
+  rm -f "$dep_chart_dir"/*.tgz
   exec_helm "dependency update $chart_dir"
-  for dep_chart_tar in "$dep_chart_dir"/*.tgz; do
-    $TAR -xf "$dep_chart_tar" -C "$dep_chart_dir"
-    $RM -f "$dep_chart_tar"
-  done
 }
 
 build_images() {


scripts/release.sh
```diff
@@ -431,15 +431,10 @@ build_helm_deps() {
   local chart_dir=${HELM_CHART_DIR%/}
   local dep_chart_dir="$chart_dir/charts"
 
-  # This performs a dependency update and then extracts the tarballs pulled.
-  # If and when the `--untar` functionality is added to the `helm dependency
-  # update command, the for block can be removed in favour of the `--untar` option.
-  # Ref: https://github.com/helm/helm/issues/8479
+  # Helm dependency update doesn't work correctly if a version of a dependency is newer than the required version
+  # This rm command clears out existing chart tarballs, so that helm dependency update works correctly.
+  rm -f "$dep_chart_dir"/*.tgz
```
tiagolobocastro, line 436: This causes unnecessary rebuilds when code has not been modified.
Here's an alternative to dependency update:

```bash
helm_dep_update_required() {
  local chart="$1"

  repository=$(echo "$chart" | jq -r '.repository')
  version=$(echo "$chart" | jq -r '.version')
  name=$(echo "$chart" | jq -r '.name')
  tar=$(echo "$chart" | jq -r '.tar')

  if [ "$(semver validate "$version")" != "valid" ]; then
    log_fatal "Found $name with version $version only pinned versions are supported!"
  fi

  if [ -z "$repository" ]; then
    echo "false"
    return 0
  fi

  if ! [ -f "$tar" ]; then
    echo "true"
  else
    echo "false"
  fi
}

# Returns the json for the helm chart dependencies
# In case of dependents of other local charts, an additional parent is added to the json
helm_all_deps() {
  local chart_rel="${1:-}"
  local chart_dir="$CHART_DIR"
  if [ -n "$chart_rel" ]; then
    chart_dir="$chart_dir/$chart_rel"
  fi
  local all_deps deps

  if ! deps=$($HELM show chart "$chart_dir" --kubeconfig "$CHART_DIR/fake" | yq -ojson '.dependencies[]' | jq -c); then
    log_fatal "Can't find the helm dependencies in $chart_dir"
  fi

  for chart in ${deps[@]}; do
    repository=$(echo "$chart" | jq -r '.repository')
    name=$(echo "$chart" | jq -r '.name')
    version=$(echo "$chart" | jq -r '.version')

    local name_rel="charts/$name"
    if [ -n "${repository:-}" ]; then
      local chart_tar="$chart_dir/$name_rel-$version.tgz"
      chart=$(echo "$chart" | jq -c ".tar = \"$chart_tar\"")

      if [ -n "$chart_rel" ]; then
        local chart_loc="$chart_dir"
        chart=$(echo "$chart" | jq -c ".chart = \"$chart_loc\"")
      fi

      if [ -n "${all_deps:-}" ]; then
        all_deps="$all_deps
        $chart"
      else
        all_deps="$chart"
      fi
      continue
    fi

    if [ -n "$chart_rel" ]; then
        name_rel="$chart_rel/$name_rel"
    fi
    # if there's no repository, it's a local chart, so check its own dependencies
    deps=$(helm_all_deps "$name_rel")
    if [ -n "${all_deps:-}" ]; then
      all_deps="$all_deps
      $deps"
    else
      all_deps="$deps"
    fi
  done

  echo "$all_deps"
}

# This fetches the dependencies in an exact version from the Chart.yaml
# NOTE: This won't work if we ever modify the Chart.yaml to specify non-pinned versions, ex: 14 vs 14.0.0
# Update can be forced with global var DEP_UPDATE="true".
helm_dep_update() {
  local update="false"

  deps=$(helm_all_deps)

  if [ "$DEP_UPDATE" = "true" ]; then
    update="true"
  else
    while read -r chart; do
      update_required=$(helm_dep_update_required "$chart")
      if [ "$update_required" = "true" ]; then
        update="true"
        break
      fi
    done <<< "$deps"
  fi

  if [ "$update" = "true" ]; then
    $HELM dependency update "$CHART_DIR" --kubeconfig "$CHART_DIR/fake"
    while read -r chart; do
      chart_loc=$(echo "$chart" | jq -r '.chart')
      if [ -n "$chart_loc" ] && [ "$chart_loc" != "null" ]; then
        $HELM dependency update "$chart_loc" --kubeconfig "$CHART_DIR/fake"
      fi
    done <<< "$deps"
  fi
}
```

.gitignore
```diff
@@ -15,6 +15,6 @@ Cargo.lock
 /package.json
 /rust-git-version
 /rust-toolchain.toml
-.idea
+**/.idea
```
tiagolobocastro, line 18: what's this change for?

tiagolobocastro:
Replaced by #142

PR #132, opened 2025-08-26
proto changes for pool grow

abhilashshetty04:


.github/workflows/pr-commitlint.yml
```diff
@@ -14,6 +14,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - name: Configure npm for public registry
+        run: npm config set registry https://registry.npmjs.org/
```
tiagolobocastro: what's this for?

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -17,7 +17,7 @@ service PoolRpc {
   rpc ExportPool (ExportPoolRequest) returns (google.protobuf.Empty) {}
   rpc ImportPool (ImportPoolRequest) returns (Pool) {}
   rpc ListPools (ListPoolOptions) returns (ListPoolsResponse) {}
-  rpc GrowPool (GrowPoolRequest) returns (GrowPoolResponse) {}
+  rpc GrowPool (GrowPoolRequest) returns (Pool) {}
```
tiagolobocastro: This is a breaking change

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -59,23 +59,18 @@ message CreatePoolRequest {
...
+  // User can pass Absolute size. For example, 800GiB, 3.5TiB or 4563726394B.
   //
-  // Values smaller that 1.0 would decrease amount of space reserved for metadata.
-  // For example, 0.5 would half the capacity reserved for metadata.
-  //
-  // The actual amount of disk space in bytes reserved for metadata may
-  // depend on pool size, cluster size, and other pool parameters.
-  // A pool backend may also allow dynamic metadata allocation.
-   optional float md_resv_ratio = 1;
+  optional string max_growth = 1;
```
tiagolobocastro: this is a breaking change

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -180,14 +177,3 @@ message GrowPoolRequest {
...
   google.protobuf.StringValue uuid = 2;   // optional uuid for the pool to be grown
 }
-
-// Grow pool response.
-message GrowPoolResponse {
-  // Pool's state before expansion.
-  Pool previous_pool = 1;
-  // Pool's state after expansion.
-  // The difference between previous_pool and current_pool can be used
-  // to determine how much the pool has grown, and if the growth
```
tiagolobocastro: One thing we miss out by removing this, is determining how much the pool has grown for example.

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -150,6 +145,8 @@ message PoolMetadataInfo {
   // existing replicas.
   // If the pool backend does not support metadata capacity evaluation, zero is returned.
   uint64 md_used_pages = 3;
+  // Maximum capacity upto which this pool can be grown, in bytes.
+  uint64 max_growable_size = 4;
```
```suggestion
  optional uint64 max_growable_size = 4;
```

apis/io-engine/protobuf/v1/pool.proto
dsharma-dc: > One thing we miss out by removing this, is determining how much the pool has grown for example.

But this will be always set to `previous` one ? which means if pool has been expanded multiple times we won't be able to figure that out anyway.

dsharma-dc: but anyway it's fine to retain this field.

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -59,23 +60,20 @@ message CreatePoolRequest {
...
-  // 1.0 is the default value, and it means to reserve the amount
-  // of disk space for metadata which is default for the pool type.
+  // 1. Using factor
+  // 1x will allow limited growth. pool cannot grow upto 2x the original size.
+  // 5x will allow the pool to be grown 5 times the initial size.
   //
-  // Values larger that 1.0 would increase amount of space reserved for metadata.
-  // For example, 2.0 would double the capacity reserved for metadata.
+  // 2. Absolute size
+  // User can pass Absolute size. For example, 800GiB, 3.5TiB or 4563726394B.
```
dsharma-dc, line 74: Do we safeguard early-on somewhere that the absolute size provided by user is indeed larger than the current disk size ? If not, that would be a good check to do.

abhilashshetty04: We can do that in io-engine after we create bdev and convert size in terms of factor.

dsharma-dc: please have a task created for that so we can handle in a different PR.

dsharma-dc (approved):

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -59,23 +60,20 @@ message CreatePoolRequest {
 
 // Pool metadata arguments.
 message PoolMetadataArgs {
-  // This parameter controls the initial metadata reservation for a storage pool.
-  // This parameter sets a ratio of the desired metadata reservation amount
-  // to the default reservation for the given pool type.
+  // Deprecated after the addition of max_expansion.
+  optional float md_resv_ratio = 1;
```
tiagolobocastro, line 64: You could also add a `PoolMetadataArgsV2` btw, and use `GrowPoolRequestV2`

tiagolobocastro (approved):

apis/io-engine/protobuf/v1/pool.proto
abhilashshetty04: GrowPoolRequest doesn't have any changes. So i guess we can reuse that in GrowPoolV2.

If we add a new message `PoolMetadataArgsV2`, the existing `PoolMetadataArgs` will still be there in `CreatePoolRequest`. Doesn't add much value i think.

apis/io-engine/protobuf/v1/pool.proto
abhilashshetty04: I have added this change in the existing PR. ptal

https://github.com/openebs/mayastor/pull/1903/files#diff-a83bdfbc4441eaa99f126846708935e036f7cc8af86f3ab85a5b7c5fb80d8b79R448

tiagolobocastro:
@mergify backport release/2.9

PR #144, opened 2025-10-28
Improve release.sh to auto handle external parent root

tiagolobocastro:
build: try to fetch parent root dir
    
    When the release.sh is sourced from another repo, the parent root may not
    be a dir away from the script dirname.
    Instead we figure out by trying well known env variables and even using
    the git top-level.
    
    Adds a debug-slim for easy quick builds in debug mode for testing.
    
---

    build: correct error msg and use type -P
    
    Correct the error message when NIX_SOURCES is not set
    Use the bash builtin to fetch the binary path


Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro:
@mergify queue

PR #146, opened 2025-11-07
build: auto-add jq to the release.sh

tiagolobocastro:
Adding jq as a dependency broke existing infra pipeline which did not use the nix-shell as a way of setting all deps.
To avoid this we simply auto-add jq if missing.


niladrih (approved):

tiagolobocastro:
@mergify queue

tiagolobocastro:
@mergify backport release/2.10

PR #150, opened 2026-01-28
feat(pool/err): add pool error api

tiagolobocastro:


apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -191,3 +202,71 @@ message GrowPoolResponse {
...
+  // Disk IO stalls
+  bool               io_stalled = 4;
+  uint64 stall_transition_count = 5;
+  uint64        stall_threshold = 6;
+}
+
+message PoolAlerts {
+  optional PoolAlertStatus status = 1;
+
+  repeated PoolAlert       notice = 2; // Does not raise any status changes...
```
Abhinandan-Purkait: Generally a normal alert or non alert is Ok or Clear is what I have seen

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -191,3 +202,71 @@ message GrowPoolResponse {
...
+  repeated PoolAlert    attention = 3;
+  repeated PoolAlert      warning = 4; // Pool becomes SUSPECTED once we have warnings
+  repeated PoolAlert     critical = 5; // Critical Pool should not be used
+}
+
+enum PoolAlertStatus {
+  Healthy   = 0; // No Alerts
+  Attention = 1; // IE some pool errors
+  Warning   = 2; // IE if pool error is above threshold || or stalled flaps above threshold
```
Abhinandan-Purkait: So we never go to critical if errors go way above threshold? Just wondering if it makes sense to have a critical threshold internally, i.e 70% above decided threshold or something?

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -191,3 +202,71 @@ message GrowPoolResponse {
...
+  Critical  = 3; // IE if IO is stalled
+}
+
+enum PoolAlert {
+  // Unrecognized
+  Unknown             = 0;
```
Abhinandan-Purkait: What is unknown is it the default for no alerts? Should it called something else?

Abhinandan-Purkait (approved):

apis/io-engine/protobuf/v1/pool.proto
tiagolobocastro: The problem is we stop using critical pools, so I think it's hard to make this decision based on the errors only
I was wondering if we'd need a bit more information, such as seen errors only go up with no completed IOs

apis/io-engine/protobuf/v1/pool.proto
tiagolobocastro: This means we don't recognize an alert, so if there's different versions talking to each for example, then we could set alert to `Unknown`.

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -96,6 +102,7 @@ enum PoolState {
   POOL_ONLINE = 1;   // the pool is in normal working order
   POOL_DEGRADED = 2; // the pool has experienced a failure but can still function
   POOL_FAULTED = 3;  // the pool is completely inaccessible
+  POOL_SUSPECTED = 4; // the pool is suspected to have some issues due to warning alerts
```
abhilashshetty04, line 104: critical alerts right?

apis/io-engine/protobuf/v1/pool.proto
abhilashshetty04: During ongoing IOStall this will be marked as Critical

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -191,3 +202,74 @@ message GrowPoolResponse {
...
+  repeated PoolAlert       notice = 2; // Does not raise any status changes...
+  repeated PoolAlert    attention = 3;
+  repeated PoolAlert      warning = 4; // Pool becomes SUSPECTED once we have warnings
+  repeated PoolAlert     critical = 5; // Critical Pool should not be used
+}
+
+enum PoolAlertStatus {
+  Healthy   = 0; // No Alerts
+  Attention = 1; // IE some pool errors
+  Warning   = 2; // IE if pool error is above threshold || or stalled flaps above threshold
```
abhilashshetty04:
```suggestion
  Warning   = 2; // IE if pool error is above threshold || or stall_tranistion within interval is above threshold
```

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -191,3 +202,74 @@ message GrowPoolResponse {
...
+  Critical  = 3; // IE if IO is stalled
+}
+
+enum PoolAlert {
+  // Unrecognized
+  Unknown             = 0;
+  // Pool IO is stalled
+  IoStalled           = 1;
+  // IO is not stalled, but it has often been stalled recently
```
abhilashshetty04: We will use this when stall_threshold is exceed within interval?

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -191,3 +202,74 @@ message GrowPoolResponse {
...
+  // Errors seen for this disk
+  PoolErrors errors = 3;
+}
+
+message ClearErrorRequest {
+  // Name of the pool
+  string          name = 1;
+  // Uuid of the pool
+  optional string uuid = 2;
+
```
abhilashshetty04, line 277: new line not needed..?

apis/io-engine/protobuf/v1/pool.proto
tiagolobocastro: warnings | critical

apis/io-engine/protobuf/v1/pool.proto
tiagolobocastro: Yes

apis/io-engine/protobuf/v1/pool.proto
tiagolobocastro: Just for formatting, keeping the name/uuid separate

apis/io-engine/protobuf/v1/pool.proto
abhilashshetty04: Warning as well...Okay

apis/io-engine/protobuf/v1/pool.proto
tiagolobocastro: I brushed up this entire enum, PTAL

abhilashshetty04 (approved):

tiagolobocastro:
@mergify queue

PR #156, opened 2026-05-10
build: drop ring feature and dep

tiagolobocastro:
Drops the ring dependency altogether.
Ensures kube crates come from workspace so we can specify the actual features on the workspace.

### DIFF
--- a/.github/workflows/pr-rust-lint.yaml
+++ b/.github/workflows/pr-rust-lint.yaml
@@ -18,4 +18,4 @@ jobs:
       - name: cargo-clippy
         run: nix-shell --pure --run ./scripts/rust/linter.sh
       - name: cargo-fmt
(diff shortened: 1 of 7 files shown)


Abhinandan-Purkait:
Whats the difference here? Is it because ring had those license issues flagged on FOSSA?

Abhinandan-Purkait:
Also, interesting the rls-aws-lc-rs seems FIPS compliant.

tiagolobocastro:
> Whats the difference here? Is it because ring had those license issues flagged on FOSSA?

No, it's help us use FIPS algorithms ; this is not yet there, we'd have to enable fips feature flag, but this is the start

niladrih (approved):

tiagolobocastro:
@mergify queue

Abhinandan-Purkait (approved):

tiagolobocastro:
@mergify queue

PR #157, opened 2026-05-31
feat: add pool_name label to replica stats

cmrmahesh:
Part of series of changes to add pool_name to `ReplicaIoStats`

Next changes in 
- `io-engine` populate `pool_name` from the replica's backing pool
- control-plane - submodule bump to pick up new proto
- mayastor-extensions — emit `pool_name` as a Prometheus label on replica I/O metrics

Motivation and Context:

Described at [#1990](https://github.com/openebs/mayastor/issues/1990)

Initial map based approach at https://github.com/openebs/mayastor-extensions/pull/860

### DIFF
--- a/apis/io-engine/protobuf/v1/stats.proto
+++ b/apis/io-engine/protobuf/v1/stats.proto
@@ -40,6 +40,8 @@ message ReplicaIoStatsResponse {
 message ReplicaIoStats {
   optional string entity_id = 1; // Entity ID of the replica, If any.
   IoStats stats = 2; // IO stats of the replica.
+  optional string poolname = 3; // Name of the pool hosting this replica.
+  optional string pooluuid = 4; // UUID of the pool hosting this replica.
 }
 
 // Common IO stats struct for all resource types.


cmrmahesh:
@tiagolobocastro  Should I add both `pool_name` and `pool_uuid`? Since we're touching the proto anyway, adding both is trivial and keeps it consistent with the `Replica` message which already carries both fields.

apis/io-engine/protobuf/v1/stats.proto
```diff
@@ -40,6 +40,7 @@ message ReplicaIoStatsResponse {
 message ReplicaIoStats {
   optional string entity_id = 1; // Entity ID of the replica, If any.
   IoStats stats = 2; // IO stats of the replica.
+  optional string pool_name = 3; // Name of the pool hosting this replica.
```
cmrmahesh: Going by the existing naming conventions I think `poolname` is better?
https://github.com/openebs/mayastor-dependencies/blob/develop/apis/io-engine/protobuf/v1/replica.proto#L51

tiagolobocastro:
> @tiagolobocastro Should I add both `pool_name` and `pool_uuid`? Since we're touching the proto anyway, adding both is trivial and keeps it consistent with the `Replica` message which already carries both fields.

Yeah I think both the pool name and the pool uuid would make sense for consistency

apis/io-engine/protobuf/v1/stats.proto
tiagolobocastro: I prefer `pool_name`, but yeah maybe `poolname` for consistency

cmrmahesh: Done

tiagolobocastro (approved):
Could you please squash the commits?

niladrih (approved):

tiagolobocastro:
@mergify queue

PR #162, opened 2026-06-22
feat(nexus): add read_only field for ROX access mode

yugchaudhari:
## Summary

Adds a `read_only` field to `PublishNexusRequest` and the `Nexus` query message in the io-engine v1 proto. First piece of the ROX-fs implementation per OEP openebs/openebs#4242.

## What

- `PublishNexusRequest.read_only` (field 5, `optional bool`): when true, the target rejects write I/O at submit time and skips acquiring write-exclusive reservations on the nexus children.
- `Nexus.read_only` (field 12, `optional bool`): reflects the currently-published state so the control plane can read it back.

Fully backwards-compatible (field is optional).

Scoped per publish, flipping between ROX and RWO is unshare + re-share with a different value, so no new mutation RPC is needed.

## Why

ROX (`MultiNodeReaderOnly`) needs N initiators to share a single read-only nexus. Write-exclusive reservations conflict with multi-host attach by construction, and the OEP calls for target-level write rejection as defense-in-depth against misconfigured mounts.

This PR is proto-only. The implementation that consumes the field will be added in openebs/mayastor (follow-up PR after this merges and the submodule bumps).

## Tracking

- OEP: openebs/openebs#4242
- OEP PR: openebs/openebs#4231
- Original feature request: openebs/mayastor#1993

cc @tiagolobocastro @Abhinandan-Purkait

### DIFF
--- a/apis/io-engine/protobuf/v1/nexus.proto
+++ b/apis/io-engine/protobuf/v1/nexus.proto
@@ -232,6 +232,7 @@ message Nexus {
   repeated string allowed_hosts = 9; // host (nqn's) which are allowed to connect to the target
   optional NexusLabelVersion label_version = 10; // version of the nexus label layoyut
   optional uint64 bdev_size = 11; // the actual size of the nexus block device
+  optional bool read_only = 12; // true when the nexus is currently published as ROX and rejects writes
 }
 
 message CreateNexusResponse {
@@ -306,6 +307,11 @@ message PublishNexusRequest {
   string key = 2; // encryption key
   ShareProtocol share = 3;  // protocol used for the front end.
   repeated string allowed_hosts = 4; // host (nqn's) which are allowed to connect to the target.
+  // When true, the target rejects write I/O at submit time and skips acquiring
+  // write-exclusive reservations on the nexus children. Used for the
+  // ReadOnlyMany (ROX) access mode. Scoped to this publish; a re-publish with
+  // a different value flips the mode.
+  optional bool read_only = 5;
 }
 
 message PublishNexusResponse {


yugchaudhari:
@tiagolobocastro @Abhinandan-Purkait when you have some time, please take a look. The [openebs/mayastor](https://github.com/openebs/mayastor) consumer side (ROX iter1, openebs/openebs#4242) is ready locally and waiting on this.

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
> @tiagolobocastro @Abhinandan-Purkait when you have some time, please take a look. The [openebs/mayastor](https://github.com/openebs/mayastor) consumer side (ROX iter1, [openebs/openebs#4242](https://github.com/openebs/openebs/issues/4242)) is ready locally and waiting on this.

Sure.

apis/io-engine/protobuf/v1/nexus.proto
```diff
@@ -143,6 +143,11 @@ message CreateNexusRequest {
   string nexusInfoKey = 9; // the key to use to persist the nexus info structure
   optional NvmeReservation resvType = 10; // the reservation type to use
   NexusNvmePreemption preemptPolicy = 11; // the preemption policy to use
+  // When true, the nexus rejects write I/O at submit time and skips
+  // acquiring write-exclusive reservations on its children. Used for the
+  // ReadOnlyMany (ROX) access mode where N initiators share a single
+  // read-only nexus. Immutable for the nexus lifetime.
```
tiagolobocastro: > Immutable for the nexus lifetime.

hmm but then we won't be able to re-publish the nexus when volume is published whilst offline rebuild is ongoing?

apis/io-engine/protobuf/v1/nexus.proto
```diff
@@ -143,6 +143,11 @@ message CreateNexusRequest {
...
+  bool read_only = 12;
```
tiagolobocastro: make these optional please

apis/io-engine/protobuf/v1/nexus.proto
yugchaudhari: Yes, you are right. The offline rebuild promote-on-publish flow shares the nexus in-place without recreating it, so if a ROX volume goes through offline-rebuild (temp nexus needs writes → read_only=false) and then a CSI publish promotes it into ROX (read_only=true), the immutability makes that in-place transition impossible.

Two ways to handle it:

1. Add a SetReadOnly RPC now, drop the immutability wording. Keeps offline rebuild's promote-in-place flow working for ROX. Adds a small amount of scope to iter1 of ROX (RPC + handler).
2. Keep it immutable, and have offline rebuild's promote-on-publish fall through to tear-down-and-recreate when the read_only setting needs to flip. Preserves the guarantee, more CP-side complexity.

I am leaning toward first option since we'll likely want runtime read_only control anyway (kubectl-mayastor toggle, or future scenarios where an admin flips a live volume between ROX and RWO without recreating). Any suggestions, which could be better?

tiagolobocastro: How about just adding the readOnly to the publish? Then you can select at publish time if you want to change or not

yugchaudhari: Made the change: dropped `read_only` from `CreateNexusRequest` and added `read_only` to `PublishNexusRequest`. Kept `read_only` on the Nexus query message so callers can see the current effective state. Flipping ROX ↔ RWO becomes unshare + re-share with different value, no new mutation RPC needed.

tiagolobocastro (approved):

yugchaudhari:
@tiagolobocastro let's merge this if it looks good :)

tiagolobocastro:
@mergify queue

PR #168, opened 2026-07-01
extending pool clear error api

abhilashshetty04:
Currently, clear error api has `Clear` enum which has `ClearAll` variant.

With this change user can selectively clear error or stall transition counts also.

### DIFF
--- a/apis/io-engine/protobuf/v1/pool.proto
+++ b/apis/io-engine/protobuf/v1/pool.proto
@@ -286,9 +286,14 @@ message ClearErrorRequest {
 }
 
 enum ClearErrors {
-  // Clears all counted errors and related alerts
-  // Example, it clears the io stall transitions, but doesn't clear an io stall
+  // Clears all counted errors and stall transitions
+  // Note: It doesn't clear an io stall state if the pool is currently in a stalled state.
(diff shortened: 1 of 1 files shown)


apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -286,9 +286,13 @@ message ClearErrorRequest {
 }
 
 enum ClearErrors {
-  // Clears all counted errors and related alerts
+  // Clears all counted errors and stall transitions
   // Example, it clears the io stall transitions, but doesn't clear an io stall
   ClearAll = 0;
+  // Clears only the counted I/O errors
+  ClearIoError = 1;
```
abhilashshetty04: Should i name this as IoError and IoStallTransitions?

Abhinandan-Purkait: I guess lets keep it similar, as the above calls it as ClearAll

Abhinandan-Purkait (approved):

apis/io-engine/protobuf/v1/pool.proto
```suggestion
  ClearIoErrors = 1;
```

tiagolobocastro (approved):

abhilashshetty04:
@Mergifyio queue

PR #171, opened 2026-07-03
feat(proto): add GetPoolHealth RPC to io-engine pool API

susobhandey:
Add a GetPoolHealth RPC to the v1 PoolRpc service so callers can query SMART/health for the disk(s) backing a pool.

Also register the new message types in the hand-maintained v1 re-export list so downstream crates (io-engine-api, control-plane rpc) pick them up.


apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -182,6 +185,44 @@ message ListPoolsResponse {
...
+  google.protobuf.StringValue uuid = 2;   // optional uuid for the pool
+}
+
+// SMART / health information for a backing device. NVMe (when attached via
+// uring/aio rather than VFIO), SATA and SAS are all read with smartctl. Fields
+// that a given device or transport cannot report are left unset.
+message DeviceHealth {
+  uint32 critical_warning = 1;                       // raw NVMe critical-warning flags (0 == none)
+  bool healthy = 2;                                  // true when no critical condition is reported
+  optional sint32 temperature_celsius = 3;
```
Abhinandan-Purkait, line 243: maybe use K, and that way we won't need to deal with -ve numbers in general.

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -182,6 +185,77 @@ message ListPoolsResponse {
...
+  string name = 1;                        // name of the pool
+  google.protobuf.StringValue uuid = 2;   // optional uuid for the pool
+}
+
+// Device identity/inventory data. Static for the device's lifetime, so
+// implementations should fetch this once and cache it rather than re-reading
+// it on every health poll -- unlike DeviceHealth's counters, which are always
+// read fresh. Fields a given device or transport cannot report are left unset.
+message DeviceIdentity {
+  optional string model = 1;
```
tiagolobocastro, line 198: Are these all optional because smart may not provide any info?

abhilashshetty04 (approved):

apis/io-engine/protobuf/v1/pool.proto
```diff
@@ -182,6 +185,77 @@ message ListPoolsResponse {
   repeated Pool pools = 1;  // list of the pools
 }
 
+message GetPoolHealthRequest {
+  string name = 1;                        // name of the pool
```
abhilashshetty04, line 189: Shall we make this optional like 

https://github.com/openebs/mayastor-dependencies/blob/develop/apis/io-engine/protobuf/v1/pool.proto#L185

If name is not provided then we return for all pools in this node

Abhinandan-Purkait (approved):

